PR #3, opened 2015-07-14
Fix issues with systemd images

nghiant2710:


systemd/Dockerfile.wheezy.tpl
```diff
@@ -0,0 +1,32 @@
+FROM #{FROM}
+
+# Install Systemd
+
+RUN echo "deb http://ftp.debian.org/debian/ wheezy-backports main contrib non-free" >> /etc/apt/sources.list
+RUN apt-get update && apt-get -t wheezy-backports install -y systemd \
```
Page-: What is `-t`?

systemd/generate-dockerfile.sh
```diff
@@ -19,8 +20,20 @@ for repo in $repos; do
 	for suite in $suites; do
 		dockerfilePath=$repo/$suite
 		mkdir -p $dockerfilePath
-		sed -e s~#{FROM}~resin/$baseImage:$suite~g Dockerfile.tpl > $dockerfilePath/Dockerfile
-		cp entry.sh launch.service $dockerfilePath/
+		
+		if [ $suite == 'wheezy' ] && [ $repo != 'rpi' ]; then
```
Page-: Can we have comments explaining why the special cases

systemd/Dockerfile.wheezy.tpl
nghiant2710: -t to retrieve packages from wheezy-backports repository. http://manpages.ubuntu.com/manpages/lucid/man8/apt-get.8.html

device-base/generate-dockerfile.sh
```diff
@@ -21,21 +35,34 @@ for device in $devices; do
...
 
-	# Only for armv7hf devices
-	if [ $device == 'raspberrypi2' ] || [ $device == 'beaglebone' ]; then
-		mkdir -p $dockerfilePath/sid
-		sed -e s~#{FROM}~resin/$baseImage:sid~g \
-			-e s~#{SUITE}~sid~g $template > $dockerfilePath/sid/Dockerfile		
-	fi
+			case "$suite" in
+			'wheezy')
+				sourcelist="$bb_sourceslist_wheezy_cmd"
```
Page-: You can use:
```
sourcelist=bb_sourceslist_$suite_cmd
${!sourcelist}
```

systemd/Dockerfile.wheezy.tpl
Page-: Can you use `--target-release` please

device-base/generate-dockerfile.sh
nghiant2710: @Page- ,this doesn't work for me

systemd/Dockerfile.wheezy.tpl
nghiant2710: Updated!

device-base/generate-dockerfile.sh
Page-: What did you try? You would have to replace the `$sourcelist` uses with `${!sourcelist}` - it's not a big deal though as I'm not sure dropping the switch to use that style would make the code any clearer anyway

nghiant2710: Oops, I get it wrong, it works. Btw I agree that this change won't make it look better.

PR #5, opened 2015-07-24
Stop installing systemd on wheezy images

nghiant2710:


nghiant2710:
@Page- , can you review this PR please?

Page-:
Why is this necessary?

nghiant2710:
@Page- http://jenkins.dev.resin.io/view/Resin%20Library/job/resin-beaglebone-buildpack-deps/ws/bashbrew/logs/build-2015-07-23--18-52-13/build-beaglebone-buildpack-deps%3Awheezy.log/*view*/

Installing systemd from backports also installs some dependencies from backports which causes unmet dependencies when installing new packages. we've agreed that if systemd doesn't play well on wheezy then we can drop it. Hence, it's better to remove it from wheezy

Page-:
Ok, if @petrosagg wants to look into this further it'd be easy enough to revert this and do the steps to fix it

PR #8, opened 2015-08-27
Fix issue with DBus on host/container

nghiant2710:


nghiant2710:
@petrosagg , @lorenzo-stoakes, @lifeeth. Can you have a look?

systemd/armv7hf/jessie/entry.sh
```diff
@@ -9,10 +9,17 @@ if [ "$INITSYSTEM" = "on" ]; then
 	echo -e "${GREEN}Systemd init system enabled."
 	env > /etc/docker.env
 
-	echo -e "#!/bin/bash\n exec $@" > /etc/resinApp.cmd
+	cat <<-EOF > /etc/resinApp.cmd
```
Page-: Can you change `.cmd` to `.sh`, since `.cmd` is a windows thing whereas `.sh` is usually used for bash scripts (or no extension at all is used at times as well)

Page-: Actually it seems unchanged from before, it just `cat`s a heredoc instead, so I guess it can stay as is

Page-:
Overall this looks ok to me, but I'm also not really sure what I'm looking at, so it'd be good if @petrosagg could take a look

PR #20, opened 2015-09-30
necessary packages for python images

nghiant2710:


python/nitrogen6x/jessie/Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		python-dev \
 		python-dbus \
 		python-pip \
+		python-virtualenv \
+		python-setuptools \
 	&& rm -rf /var/lib/apt/lists/*
```
emirotin: are all of these files generated from the template file?

emirotin:
Assuming the answer yes LGTM

petrosagg:
@emirotin yes. For these type of PRs it's better to check commit by commit as there is one that has the template changes and then another one with all the autogenerated files from the templates.

petrosagg:
SGTM

PR #21, opened 2015-09-30
Golang images

nghiant2710:


golang/generate-dockerfile.sh
```diff
@@ -0,0 +1,102 @@
...
+	'beaglebone')
+		binary_url=$resinUrl
+		binary_arch='armv7hf'
+	;;
+	'edison')
+		binary_url=$nodejsUrl
+		binary_arch='386'
+	;;
+	'nuc')
+		binary_url=$nodejsUrl
```
lifeeth: There is no nodejs URL here.

nghiant2710: @lifeeth updated!
It's should be golangUrl

PR #24, opened 2015-10-16
Set hostname to <device-type>-<device uuid> and update wrong link

nghiant2710:


shaunmulligan:
LGTM, and sorry for not pinging you about changing the structure of the docs and breaking links, my bad :/

shaunmulligan:
the better way would be to use %%RESIN_MACHINE_NAME%% but that would require switching to dockerfile.template and not sure if we want that or if it would cause problems?

nghiant2710:
entry.sh script will be called on device so I don't think we have %%RESIN_MACHINE_NAME%% available

shaunmulligan:
sorry what i meant was you could set ENV deviceName = %%RESIN_MACHINE_TYPE%%, this would work at least until we add device_name and other Resin variables to the device

nghiant2710:
oh yes, makes sense

shaunmulligan:
Are there any obvious issues with switching to dockerfile.template and doing it this way @nghiant2710 and @petrosagg ?

PR #26, opened 2015-10-19
Set hostname to <device-type>-<device uuid>

nghiant2710:


shaunmulligan:
LGTM as a temp solution +1

PR #27, opened 2015-10-22
Remake Python images (Install Python from source). Add Python3 images

nghiant2710:


petrosagg:
If we're going to compile python from source you need to verify the origin of the packages using the GPG of the author. See how here https://github.com/docker-library/python/blob/master/2.7/Dockerfile

nghiant2710:
@petrosagg Updated! I've added gpg verification.

PR #28, opened 2015-11-02
Remake Python images: install from source, add Python3 images.

nghiant2710:
- Install from source.
- Add Python3 images.


nghiant2710:
@petrosagg all conflicts resolved

PR #33, opened 2015-11-07
Beaglebone black gpg fix

petrosagg:
Previously trusted gpg keys were downloaded over http during build time.


nghiant2710:
@petrosagg These keys are not available on key server, that's why I have to download them directly. Refer here: https://groups.google.com/forum/#!category-topic/beagleboard/ubuntu/Qfv6gpqYh5s.

petrosagg:
I just uploaded them, http://keyserver.ubuntu.com/pks/lookup?op=vindex&search=0xD284E608A4C46402&fingerprint=on

nghiant2710:
SGTM

PR #37, opened 2015-11-10
Fix libmraa issues

nghiant2710:


device-base/Dockerfile.i386.edison.tpl
```diff
@@ -1,29 +1,42 @@
...
 		i2c-tools \
 		iputils-ping \		
 		ifupdown \				
-		python-dev \
-		swig \
 		usbutils \		
 	&& rm -rf /var/lib/apt/lists/*
 
+# MRAA v0.8.1 commit 
+ENV MRAA_COMMIT 049ba5fa9f2d18ac0ec6729c46916b34998d3c5f
```
Page-: You can use `ENV MRAA_COMMIT v0.8.1` and it will check out based on the tag

nghiant2710: Petros wants to use the commit hash so that it can't be fake

device-base/edison/jessie/Dockerfile
```diff
@@ -1,29 +1,42 @@
...
 		usbutils \		
 	&& rm -rf /var/lib/apt/lists/*
 
+# MRAA v0.8.1 commit 
+ENV MRAA_COMMIT 049ba5fa9f2d18ac0ec6729c46916b34998d3c5f
+
 # Install mraa
-RUN git clone https://github.com/intel-iot-devkit/mraa.git \
-	&& mkdir mraa/build && cd mraa/build \
+RUN set -x \
```
Page-: Why `set -x`?

nghiant2710: the reason is Petros wants to print the commands to the build logs

PR #40, opened 2015-11-12
 Use /dev/console as the TTY device path.

lorenzo-stoakes:
In some cases it seems /dev/pty/0 is not populated. If it isn't and we have TTYPath=/dev/pts/0 (as it previously was set), the launch.service service will fail to run with an error and the user's specified CMD line in their Dockerfile simply won't run.
    
/dev/console should always be available, so this is a safe alternative, and is also an appropriate TTY to use.


nghiant2710:
LGTM

PR #43, opened 2015-11-16
Fix issue with input format of old version sha256sum

nghiant2710:
The old version of sha256sum (v8.13 which is on wheezy images) wants the exact format for the input. It needs 2 whitespaces.


nghiant2710:
@petrosagg, @Page- , Can you review this?

PR #45, opened 2015-11-18
 Mount devtmpfs _over_ /dev/pts.

lorenzo-stoakes:
Failing to do so results in devtmpfs's /dev/ptmx causing /dev/ttyp[0-f] files to be used rather than /dev/pts/<fd> and if we then move /dev/pts back into place after mounting devtmpfs, libc gets confused and tries to get /dev/ptmx allocated files in /dev/pts/<fd>, causing forkpty calls (+ anything that tries to look up the pty path) to fail.
    
Alternative working solutions are to symlink or loop mount /dev/pts/ptmx to /dev/ptmx (this doesn't seem to retain /dev/pts/[0-2]), or to unmount /dev/pts (this results in use of /dev/ttyp[0-f] rather than /dev/pts/<fd> however.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 2015-11-18
+
+### Changed
+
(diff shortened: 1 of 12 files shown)


lorenzo-stoakes:
This is confused, am withdrawing to regroup it later.

PR #49, opened 2015-12-07
Add new features SSH_MODE, SYNC_MODE

nghiant2710:
This feature will serve resin sync.
Automatically setting up ssh on the container.
Enable SSH_MODE by setting `SSH_MODE on` as environment variable.
Optional Configurations:
Change the configuration by setting new value in environment variable, otherwise default value will be used.
- SSH_PASSWORD: password for ssh. Default is `resin`.
- RESIN_API_KEY: Resin API key (can be generated using Resin CLI or Resin SDK). If Resin API key is set, ssh keys added to your Resin account will be downloaded and added to the authorized_keys in the container.
- AUTH_TOKEN: Resin auth token (on the dashboard in `Preferences` section) This token will expire in short period of time so it should be used with caution.If Auth token is set, ssh key added to your Resin account will be downloaded and added to the authorized_keys in the container.
- SSH_PORT: ssh port. Default is `22`.

[Review on Reviewable](https://reviewable.io/reviews/resin-io-library/base-images/49)


lifeeth:
I have a few questions:

1. Can we use dropbear instead of openssh-server - there will be a size difference.
2. Why port 80 as default - Can we have the default as 22.

nghiant2710:
- Default port is set to 22.
- I will check if dropbear works fine in our case

nghiant2710:
New feature SYNC_MODE:
Enabling this feature and you will be able to use resin sync (resin cli plugin).
Set `SYNC_MODE on` as environment variable to turn on sync mode on base-images. `SYNC_MODE` will automatically enable `SSH_MODE` on port 80.

shaunmulligan:
@nghiant2710 how much larger does this make our *-python and *-node images? Also could you build a test base image for me to test?

nghiant2710:
@shaunmulligan, this will make `*-systemd` images 7 MB larger (dropbear, curl, ca-certificates, rsync) but the size difference will be smaller for `-python` and `-node` images since curl, ca-certificates are pre-installed.
You can test `nghiant2710/device-sync:jessie` which is based on `resin/armv7hf-systemd:jessie`, will build python and node version later.

shaunmulligan:
@nghiant2710 I think we need to add this line:
`RUN sed -i 's/UsePAM yes/UsePAM no/' /etc/ssh/sshd_config`
so that SSH will work on devices with 3.10 kernel...like the intel edison.

systemd/amd64/jessie/entry.sh
```diff
@@ -23,6 +23,16 @@ mount --move /tmp /dev
 # ref: https://www.kernel.org/doc/Documentation/filesystems/devpts.txt
 ln -sf /dev/pts/ptmx /dev/ptmx
 
+if [ "$SSH_MODE" = "on" ] || [ "$SYNC_MODE" = "on" ]; then
+	if [ "$SYNC_MODE" = "on" ]; then
```
pcarranzav: Tabulation is a bit weird here

shaunmulligan:
This PR can probably be closed now

PR #51, opened 2015-12-08
fix #50. Fix imprope CMD instruction parsing

### LINKED ISSUE #50: Improper CMD instruction parsing when using the CMD shell form
This investigation started when we noticed that [this](https://github.com/resin-io-projects/electron-rpi-quick-start) project wouldn't run as expected when installed on an Rpi2. Instead of the electron app, the application only presented a single xterminal, which is the default behavior of xinit.

Let's start by looking into the base image hierarchy. There we find the [armv7hf-debian](https://github.com/resin-io-library/base-images/blob/master/systemd/armv7hf/jessie/Dockerfile) base image. In its Dockerfile we can find this instruction:

```
ENTRYPOINT ["/usr/bin/entry.sh"]
```

Now our affected application uses a Dockerfile.template, which is parsed and results in a Dockerfile that looks like this:

```
FROM resin/raspberrypi2-node:4.0.0
[...]
CMD xinit /usr/src/app/launch_app.sh --kiosk -- -nocursor
```

The effect of the CMD command is that it passes arguments to the ENTRYPOINT command; for more information you can refer to the related *[Exec form ENTRYPOINT example](http://docs.docker.com/engine/reference/builder/#entrypoint)* in the official Docker documentation.

Also notice that in our app's Dockerfile we use the CMD shell form, so the command will execute in ```/bin/sh -c```. As a result, the command that will be executed is:

```
/usr/bin/entry.sh /usr/sh -c xinit /usr/src/app/launch_app.sh --kiosk -- -nocursor
```

The arguments are then parsed by the **entry.sh** file:

```bash
if [ "$INITSYSTEM" = "on" ]; then
[...]
  echo -e "#!/bin/bash\n exec $@" > /etc/resinApp.sh
  chmod +x /etc/resinApp.sh
[...]
```

So the final resinApp.sh on the device looks like this:

```bash
#!/bin/bash
exec /usr/sh -c xinit /usr/src/app/launch_app.sh --kiosk -- -nocursor
```

And there's the issue. ```/bin/sh -c``` expects a string wrapped in '' or "", or else it needs special care to pass the arguments after the next parameter. In our case, xinit will be executed and the rest of the parameters will simply be ignored, as we can see in the related [sh(1)](https://www.freebsd.org/cgi/man.cgi?query=sh&sektion=&n=1) section:

```
The -c option causes the commands to be read from the string operand
     instead of	from the standard input.  Keep in mind that this option	only
     accepts a single string as	its argument, hence multi-word strings must be
     quoted.
```

Now I'm not sure if this is a docker issue or ours, some possible solutions are:

1. Support only the CMD exec form (```CMD ["command, "param1", "param2"]```) and update the docs accordingly.
2. Update the scripts to handle CMD shell form, i.e wrap the command properly in "".

nghiant2710:


petrosagg:
You have to use `%q` otherwise this can happen:

```
petrosagg@rachmaninoff ~ % cat test.sh
#!/bin/bash

args=( "$@" )
echo -e "#!/bin/bash\n exec "$(printf "'%s' " "${args[@]}")
petrosagg@rachmaninoff ~ % ./test.sh test\'
#!/bin/bash
 exec 'test''
```

lekkas:
Not directly related to @nghiant2710 current changes, I'm also posting the following info for reference because it took me a while to figure out why [this](https://github.com/nghiant2710/base-images/commit/41926a7244ed40c7811557e380b8737015bb1436#diff-4ab8a5481eadfdec361d9405e5314f56R46):

```bash
CMD=$(which $1)
shift
exec "$CMD" "$@"
```
works in the no-systemd (```else```) case. The reason is that a Dockerfile CMD instruction in shell form following an ENTRYPOINT instruction (like in our case) passes __3__ separate arguments to the command set in ENTRYPOINT:

* '/bin/sh'
* '-c'
* 'what follows the CMD instruction'

nghiant2710:
@petrosagg, I've updated this PR. Please have a look,

PR #53, opened 2015-12-11
Fix #52. Systemd booting logs issue.

### LINKED ISSUE #52: systemd booting logs are showed on resin dashboard logs

nghiant2710:


shaunmulligan:
it may be cool to make this user settable, so it is by default off, but if a user wants they can set and env var and then the systemd logs will be shown...just an idea

nghiant2710:
@shaunmulligan, I don't think we should let user do that since these logs are not useful for users. They can retrieve logs for each service using `journalctl`

shaunmulligan:
They may be useful, one could see if devices are being plugged in correctly or if a service is starting up and failing etc. But i guess it is possible through journalctl, i guess maybe @petrosagg might have some feelings on this

petrosagg:
I don't mind either way. SGTM

PR #58, opened 2015-12-22
Fix #55, bad error message when the CMD points to a file which does not exist.

### LINKED ISSUE #55: Bad error message when the CMD points to a file which does not exist
When trying in our image you get
```
$ docker run --privileged --env="RESIN_DEVICE_UUID=test" -it resin/raspberrypi-systemd foobar
/usr/bin/entry.sh: line 47: exec: : not found
```
which is not very informative, when trying in sh you get:
```
$ foobar
sh: 1: foobar: not found
```
which is a lot easier to understand

nghiant2710:


nghiant2710:
@petrosagg, I've updated your comments

nghiant2710:
@petrosagg, exit code added!

petrosagg:
SGTM

PR #59, opened 2015-12-25
Update node, python, go base-images

nghiant2710:


petrosagg:
node v5.0.0 dockerfiles are still generated in the generate-all script

nghiant2710:
@petrosagg, We no longer use that script, only the latest script. I maintain that script only for future use

petrosagg:
SGTM

PR #63, opened 2016-01-05
fix #50. Fix imprope CMD instruction parsing

### LINKED ISSUE #50: Improper CMD instruction parsing when using the CMD shell form

nghiant2710:


nghiant2710:
Moving https://github.com/resin-io-library/base-images/pull/51 here

petrosagg:
Hey, can you rebase this branch on top of origin/master instead of having a merge commit in the PR?

nghiant2710:
@petrosagg, can you review this PR then I will tidy up afterward

systemd/amd64/jessie/entry.sh
```diff
@@ -28,7 +46,9 @@ if [ "$INITSYSTEM" = "on" ]; then
 	echo -e "${GREEN}Systemd init system enabled."
 	env > /etc/docker.env
 
-	echo -e "#!/bin/bash\n exec $@" > /etc/resinApp.sh
```
petrosagg: Why don't we just replace this with

```bash
printf '#!/bin/bash\n exec ' > /etc/resinApp.sh
printf '%q ' "$0" "$@" > /etc/resinApp.sh
```

instead of doing all the above?

nghiant2710: @petrosagg, Updated!

Switching to this way because it behaves exactly as how docker handles CMD command. After doing some tests, the old solution behavior is not the same as docker when we use CMD in exec form to call a shell (ex `CMD [ "sh", "-c", "echo 123 345 456\'$%^£$%^" ]`)

petrosagg:
SGTM

PR #76, opened 2016-02-05
set STOPSIGNAL=SIGTERM to make sure the user process receives correct signal

nghiant2710:


petrosagg:
Closing this as SIGTERM is the deafult signal

PR #77, opened 2016-02-16
Gracefully stop user process when container stops

nghiant2710:


nghiant2710:
@petrosagg updated your comment about trapping SIGTERM

shaunmulligan:
@nghiant2710 what is happening with this PR?

nghiant2710:
@shaunmulligan This PR waits for some changes in the supervisor.
The supervisor needs to send correct signal to the container on stop

shaunmulligan:
@nghiant2710 can you link this issue to its corresponding supervisor issue or PR?

nghiant2710:
we don't need to PR the supervisor. This PR will work if we update Docker to v1.11 so we merge when hostOS updates Docker to that version

shaunmulligan:
@nghiant2710 , @petrosagg since we can't move to docker 1.11 anytime soon, what is the alternative plan for this?

nghiant2710:
@shaunmulligan ,we should change the supervisor a bit. I created an issue in supervisor https://github.com/resin-io/resin-supervisor/issues/233.

Page-:
@nghiant2710 what feature from 1.11 does this need?

petrosagg:
@Page- I don't think we need 1.11

@nghiant2710 can we revive this PR? The `STOPSIGNAL` Dockerfile instruction was supported on docker 1.9 and our devices have docker 1.10 so we don't need any integration with the supervisor.

Let's drop the `RESIN_STOP_SIGNAL` env var and only keep `STOPSIGNAL`. You can hardcode the signal number in `entry.sh` where you install the signal handler for the non-systemd case.

nghiant2710:
@petrosagg `STOPSIGNAL` directive on docker 1.9 and 1.10 is buggy, it's fixed in docker 1.11 https://github.com/docker/docker/issues/19300.
I've not tried this on docker 1.11 yet but it didn't work for me with docker 1.9 and 1.10

Page-:
@petrosagg I didn't think so either, that's precisely why I was asking

nghiant2710:
@petrosagg I've brought this back but confirmed that STOPSIGNAL doesn't work so we should reopen this issue https://github.com/resin-io/resin-supervisor/issues/233.
Connects to https://github.com/resin-io-playground/resin-fedora/pull/5 and https://github.com/resin-io-library/resin-rpi-raspbian/pull/36

petrosagg:
I see. The patch is minimal. I think we should backport it instead of implementing it in the supervisor.

debian/aarch64/jessie/entry.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Send SIGTERM to child processes of PID 1.
+function signal_handler()
+{
+	pkill -P 1
```
petrosagg: Can't we only kill the `$pid` variable set in `init_non_systemd`?

nghiant2710: yeah, makes sense. Updated!

debian/aarch64/jessie/entry.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+# Send SIGTERM to child processes of PID 1.
+function signal_handler()
+{
+	pkill -P $pid
```
shaunmulligan: I think this should be `kill $pid` CC @petrosagg @nghiant2710

petrosagg: yep

shaunmulligan:
fixes https://github.com/resin-io/hq/issues/87

petrosagg:
SGTM

PR #78, opened 2016-02-21
Fix #71, Add checksum verification for downloads

### LINKED ISSUE #71: Add sha256 verification for downloads
In https://github.com/resin-io-library/base-images/blob/master/golang/nuc/1.5/Dockerfile#L6 (and probably other places), it would be good to check the file is valid by performing a sha256sum comparison for the download with the hash provided by the vendor, as is done in https://github.com/resin-io-library/base-images/blob/master/python/Dockerfile.tpl#L47-L51

nghiant2710:


petrosagg:
SGTM

PR #82, opened 2016-03-05
Make sure that tags will not be duplicated when generating library fi…

nghiant2710:
…le for Golang base images


petrosagg:
SGTM

PR #83, opened 2016-03-09
Drop support for Python 3.2 base image.

nghiant2710:
This version is no longer supported in docker official library, virtualenv package also drops support for python version 3.0-3.2. So we remove this version from our library


petrosagg:
SGTM

PR #84, opened 2016-03-09
	Add latest node 0.12.x to wheezy base images

nghiant2710:


node/generate-latest-dockerfile.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -e
 
+SHA256_0_12_7_x86="bccf75736b64bd175b45681ed83a020f0dcc59b3626bbcefd5f7438aed9e9c15  node-v0.12.7-linux-x86.tar.gz"
+SHA256_0_12_7_x64="6a2b3077f293d17e2a1e6dba0297f761c9e981c255a2c82f329d4173acf9b9d5  node-v0.12.7-linux-x64.tar.gz"
+SHA256_0_12_7_armel="f29f88ef13c6dd4591a8124efcc40594e7ba4780e14b3157ba7b358a191a2a64  node-v0.12.7-linux-armel.tar.gz"
+SHA256_0_12_7_armv6hf="44d1254c82336e3da6f355f3a8a70034b761ce9b068a96529ae8d3ada000f6a6  node-v0.12.7-linux-armv6hf.tar.gz"
+SHA256_0_12_7_armv7hf="5bd808a795179923a7a6fcb1a9c2bfeaa3d59cf8cbf411a8c276ec116a89a230  node-v0.12.7-linux-armv7hf.tar.gz"
```
petrosagg: Shouldn't these be in the SHASUMS256 file?

nghiant2710: @petrosagg, I want to hardcode them here to simplify the script. otherwise we'll need to add lot of unnecessary instructions just to read these checksums and replace them in the right place (check the arch, check the node version, if the node version is 0.12.x and suite is wheezy then find v0.12.7...). Since we drop support for newer node version on wheezy base images, so we won't need to update this in the future.

petrosagg:
SGTM

PR #85, opened 2016-03-10
Add support for Raspberry Pi 3

nghiant2710:


shaunmulligan:
@nghiant2710 we probably need to update the docs about all the new images and how they fit into the base images tree. But otherwise this all looks good.

nghiant2710:
Yes indeed. I will review and update all base image docs, probably next week

petrosagg:
SGTM

PR #90, opened 2016-03-22
Fix file name in library generating script for buildpack-deps base images

nghiant2710:
File name must use this scheme <device>-<distro>-buildpack-deps


petrosagg:
SGTM

PR #91, opened 2016-03-22
Some bug fixes

nghiant2710:
1. 	Omit distro in library name if it is Debian.
2. 	Remove unsupported images. Since we do not have armel alpine base image yet so generation scripts updated to not generate Dockerfiles for these images.


petrosagg:
SGTM

PR #92, opened 2016-03-23
Temporarily remove libcurl package from alpine buildpack-deps base images

nghiant2710:


petrosagg:
Why do we need libcurl in the first place?

nghiant2710:
@petrosagg it's the equivalent package with libcurl4-openssl-dev in Debian. It's a part of buildpack-deps image.

petrosagg:
SGTM

PR #93, opened 2016-03-23
Build minimum base images for architectures from official Docker images

nghiant2710:


petrosagg:
Our current base images also include the `sudo` package

nghiant2710:
Ah, right!
Will update now

nghiant2710:
@petrosagg, Updated!

petrosagg:
SGTM

PR #96, opened 2016-03-26
Fix issue with empty binary url on ts7700 and raspberrypi3

nghiant2710:


petrosagg:
I only see a change for ts7700, not rpi3

nghiant2710:
we have a list of devices and rpi3 is after ts7700. In ts7700 case, the binary url is assigned to resinURL which is empty; by removing that line, this fix will resolve issue for both @petrosagg

petrosagg:
SGTM

PR #112, opened 2016-04-13
Implement INITSYSTEM for Alpine base images.

nghiant2710:


alpine/Dockerfile.tpl
```diff
@@ -4,9 +4,24 @@ FROM #{FROM}
...
+RUN apk add --update bash udev dbus tar ca-certificates openrc \
 	&& rm -rf /var/cache/apk/*
 
+# Config OpenRC
+RUN echo 'rc_sys="lxc"' >> /etc/rc.conf \
+	&& echo 'rc_provide="loopback net"' >> /etc/rc.conf \
+	&& echo 'rc_logger="YES"' >> /etc/rc.conf \
+	&& echo 'rc_env_allow="*"' >> /etc/rc.conf \
+	&& echo 'rc_crashed_start="YES"' >> /etc/rc.conf \
+	&& echo 'rc_crashed_stop="NO"' >> /etc/rc.conf \
```
petrosagg: It makes sense to have `/etc/rc.conf` as a file in the repo and then `COPY` it in instead of echoing all the lines

nghiant2710: @petrosagg, `rc.conf` added

alpine/resin
```diff
@@ -0,0 +1,66 @@
...
+	else
+		for pid in $pids; do
+			recursive_term $pid
+		done
+	fi
+}
+
+start_pre() {
+	# Start Udev manually
+	rc-service udev start
```
petrosagg: Why do we start an other service in `start_pre`?

nghiant2710: @petrosagg udev added to default runlevel but somehow it doesn't start automatically. So we need to start it manually

alpine/resin
```diff
@@ -0,0 +1,66 @@
+#!/sbin/openrc-run
```
petrosagg: I'm a bit confused why this service is so complex. Can't we have something simpler like this? https://aur.archlinux.org/cgit/aur.git/tree/ModemManager?h=modemmanager-openrc

alpine/resin
petrosagg: That's a bug though, we can't merge this

alpine/resin
nghiant2710: @petrosagg, I've tried the start/stop templates like that but can't find the detail docs for it, I don't know how to pass the environment variables or set the working directory for the daemon using that template. that's why I used start-stop-daemon.

nghiant2710: This service is a bit complex since I need to manually clean up all the processes created by the service. It depends on how users write the CMD command in Dockerfile (shell form or exec form)  that creates many child processes while OpenRC only handles parent process (process in resinapp.pid).
I'm thinking about using process supervision like S6 https://github.com/OpenRC/openrc/blob/master/s6-guide.md to simplify our service

alpine/resin
nghiant2710: @petrosagg, I don't think this will harm our image since it only spawns the udev daemon. You can read udev init script here: http://pastebin.com/CHBK1yzp

nghiant2710: Or we can remove this and let users start it manually

nghiant2710:
@petrosagg, I just found out that udev service couldn't be started properly on edison board, so I will remove udev from `start_pre` script. Users can start it manually if they want to

alpine/rc.conf
```diff
@@ -0,0 +1,241 @@
+# Global OpenRC configuration settings
```
petrosagg: It makes sense for this file to only contain the setting we set. I had to diff it with the default one getting installed to find out which extra options got enabled

alpine/rc.conf
```diff
@@ -0,0 +1,241 @@
...
+# Example - rc_hotplug="net.wlan !net.*"
+# This allows net.wlan and any service not matching net.* to be hotplugged.
+# Example - rc_hotplug="!net.*"
+# This allows services that do not match "net.*" to be hotplugged.
+
+# rc_logger launches a logging daemon to log the entire rc process to
+# /var/log/rc.log
+# NOTE: Linux systems require the devfs service to be started before
+# logging can take place and as such cannot log the sysinit runlevel.
+rc_logger="YES"
```
petrosagg: Why do we enable this logger?

alpine/rc.conf
```diff
@@ -0,0 +1,241 @@
...
+# "uml"            - Usermode Linux
+# "vserver"        - Linux vserver
+# "systemd-nspawn" - Container created by the systemd-nspawn utility
+# "xen0"           - Xen0 Domain
+# "xenU"           - XenU Domain
+# If this is commented out, automatic detection will be used.
+#
+# This should be set to the value representing the environment this file is
+# PRESENTLY in, not the virtualization the environment is capable of.
+rc_sys="lxc"
```
petrosagg: Why not `docker`?

alpine/resin
petrosagg: openrc has manpages you can check https://github.com/OpenRC/openrc/tree/master/man

petrosagg: For example there is a `chdir` directive http://man7.org/linux/man-pages/man8/start-stop-daemon.8.html

alpine/rc.conf
```diff
@@ -0,0 +1,241 @@
...
+
+# rc_nostop is a list of services which will not stop when changing runlevels.
+# This still allows the service itself to be stopped when called directly.
+#rc_nostop=""
+
+# rc will attempt to start crashed services by default.
+# However, it will not stop them by default as that could bring down other
+# critical services.
+rc_crashed_stop=NO
+rc_crashed_start=YES
```
petrosagg: Aren't these the default settings?

alpine/resin
nghiant2710: @petrosagg, I created our init script based on `start-top-daemon` manpages, it's easy to find them. But we're talking about simpler templates like https://aur.archlinux.org/cgit/aur.git/tree/ModemManager?h=modemmanager-openrc or https://wiki.alpinelinux.org/wiki/Writing_Init_Scripts. it's `the start/stop templates, I can't find manpages for them and clueless about what are the available `special predefined variables`.

alpine/rc.conf
nghiant2710: I think a full configuration file will help our users when they want to override our settings but it makes sense if this file contains only the setting we set. Will update and add a ref to default one.

alpine/rc.conf
nghiant2710: I want to keep the logs about rc process (starting logs... ) as well not just logs of services

alpine/rc.conf
nghiant2710: specifying `docker` or `lxc` will tell openrc it's running in a container but I think it's better to follow what Gentoo does in the Dockerfile https://github.com/gentoo/gentoo-docker-images/blob/master/amd64/Dockerfile

alpine/rc.conf
nghiant2710: yup, I want to explicitly specify settings we want to have in the base image

alpine/rc.conf
petrosagg: Won't that create a file based log and write to disk?

alpine/resin
petrosagg: The special predefined variables seem to be the `start-stop-daemon` options

nghiant2710: Will give it a try

alpine/rc.conf
nghiant2710: yup, it will log to `/var/log/rc.log`

petrosagg: Ok. we generally don't want to write to the SD card. If we can't get these logs without writing files then we should just drop this.

alpine/resin
nghiant2710: @petrosagg found the manpage http://manpages.ubuntu.com/manpages/wily/man8/openrc-run.8.html. However, it doesn't work as I expect. Service status will always be starting (because we run a shell script) even when I force it into background, which is not good.

alpine/entry.sh
```diff
@@ -24,15 +24,35 @@ mount --move /tmp /dev
...
-# echo error message, when executable file doesn't exist.
-if [  $? == '0' ]; then
-	shift
-	exec "$CMD" "$@"
+	sed -i -e s~#{DIR}~"$(pwd)"~g \
+		-e s~#{ENV}~"$(env | sed '/PATH=/d'  | tr '\n' ' ' | xargs printf '--env %s ')"~g /etc/init.d/resin
+
+	rc sysinit --quiet
+	rc default 
+	tail -f /var/log/resinapp.log
```
petrosagg: This doesn't look right. PID 1 should ultimately become the openrc process. Here this script will remain as PID 1 and inherit all orphaned processes

nghiant2710: @petrosagg OpenRC won't start in a container where it's not pid 1
```
/sbin/init
init: must be run as PID 1
```
So not like systemd, we can't make it PID 1. However, we can replicate the init process by run openrc at `sysinit` runlevel to initialize the system and start services as we want.
Our service will always clean up all processes on stop or restart so there will be no orphaned processes that will potentially break the service

petrosagg: systemd also needs to run as PID 1. The problem is that you're running `/sbin/init`, which starts a child process. What you want is `exec /sbin/init` which will replace the current process with openrc, and therefore it will be PID 1

petrosagg: We do the same here https://github.com/resin-io-library/base-images/blob/master/systemd/entry.sh#L41

nghiant2710: @petrosagg aha, I completely forgot this magic. PR updated!

alpine/entry.sh
```diff
@@ -24,15 +24,34 @@ mount --move /tmp /dev
...
+if [ "$INITSYSTEM" = "on" ]; then
+	GREEN='\033[0;32m'
+	echo -e "${GREEN}OpenRC init system enabled."
+	env >> /etc/docker.env
+	printf '#!/bin/bash\n' > /etc/resinApp.sh
+	printf '%q ' "$@" >> /etc/resinApp.sh
+	chmod +x /etc/resinApp.sh
+
+	touch /var/log/resinapp.log
+	tail -f /var/log/resinapp.log &
```
petrosagg: This log file will generate SD cards writes for the logs and will grow indefinitely

nghiant2710: We don't have a better solution to forward service log atm. Probably S6 tool can be a solution but should save it for the next PR

alpine/resin
```diff
@@ -0,0 +1,25 @@
...
+
+description="Resin.io User Application"
+log_path=/var/log
+log_file=resinapp.log
```
petrosagg: These should be removed since we're using /dev/console

alpine/entry.sh
petrosagg: We don't need these two lines anymore

alpine/resin
nghiant2710: Updated!

alpine/entry.sh
nghiant2710: Removed!

alpine/entry.sh
```diff
@@ -24,15 +24,31 @@ mount --move /tmp /dev
...
-# echo error message, when executable file doesn't exist.
-if [  $? == '0' ]; then
-	shift
-	exec "$CMD" "$@"
+
+if [ "$INITSYSTEM" = "on" ]; then
+	GREEN='\033[0;32m'
+	echo -e "${GREEN}OpenRC init system enabled."
+	env >> /etc/docker.env
+	printf '#!/bin/bash\n exec ' > /etc/resinApp.sh
```
petrosagg: Can we drop `bash` as a dependency? Why do we install it?

alpine/Dockerfile.tpl
```diff
@@ -4,9 +4,18 @@ FROM #{FROM}
 
 #{QEMU}
 
-RUN apk add --update bash udev dbus tar ca-certificates \
+RUN apk add --update bash udev dbus tar ca-certificates openrc \
```
petrosagg: Why do we depend on `bash` and `tar`?

nghiant2710: @petrosagg bash is not pre-installed and tar is too old so need to be updated.

alpine/entry.sh
nghiant2710: bash is a common shell. it makes sense to install it, doesn't it?

petrosagg:
SGTM

PR #117, opened 2016-04-23
Fix name of version comparing function

nghiant2710:


node/generate-latest-dockerfile.sh
```diff
@@ -131,7 +131,7 @@ for device in $devices; do
 		# Debian.
 		# For armv7hf and armv6hf, if node version is less than 4.x.x (0.10.x 0.12.x) then that image will use binaries from resin, otherwise it will use binaries from official distribution.
```
petrosagg: Can you change this comment to reflect the code? i.e if node version is **greater or equal than 4.x.x**...

node/generate-all-dockerfile.sh
```diff
@@ -87,7 +87,7 @@ for device in $devices; do
 
 		# For armv7hf and armv6hf, if node version is less than 4.x.x (0.10.x 0.12.x) then that image will use binaries from resin, otherwise it will use binaries from official distribution.
```
petrosagg: Can you change this comment to reflect the code? i.e if node version is **greater or equal than 4.x.x**...

node/generate-latest-dockerfile.sh
nghiant2710: @petrosagg, Updated!

petrosagg:
SGTM

PR #121, opened 2016-05-09
Few updates for base images

nghiant2710:


debian/generate-dockerfile.sh
```diff
@@ -35,7 +35,7 @@ for arch in $archs; do
 	;;
 	'armel')
 		baseImage='armel/debian'
-		label='io.resin.architecture="armel" io.resin.qemu.version="'$QEMU_VERSION'"'
+		label='io.resin.architecture="armv5e" io.resin.qemu.version="'$QEMU_VERSION'"'
```
petrosagg: Why do we do this change?

device-base/Dockerfile.armv7hf.rpi.tpl
```diff
@@ -4,6 +4,13 @@ LABEL io.resin.device-type="#{DEV_TYPE}"
 
 RUN echo "deb http://archive.raspbian.org/raspbian #{SUITE} main contrib non-free rpi firmware" >>  /etc/apt/sources.list
 RUN apt-key adv --keyserver pgp.mit.edu  --recv-key 0x9165938D90FDDD2E
+RUN apt-get update && apt-get install -y --no-install-recommends wget \
+	&& wget http://archive.raspberrypi.org/debian/raspberrypi.gpg.key \
```
petrosagg: This is totally unsecure. You're downloading a key over HTTP and then trusting it.

device-base/Dockerfile.armv7hf.rpi.tpl
```diff
@@ -4,6 +4,13 @@ LABEL io.resin.device-type="#{DEV_TYPE}"
...
+	&& apt-key add raspberrypi.gpg.key \
+	&& rm -rf raspberrypi.gpg.key \
+	&& echo "deb http://archive.raspberrypi.org/debian/ jessie main" >> /etc/apt/sources.list \
```
petrosagg: Is this the debian repo? It doesn't look like it has armv6 packages and mixing debian and raspbian doesn't look like a good idea

petrosagg:
There is some merge commit in the history, probably needs a rebase

device-base/Dockerfile.armv7hf.rpi.tpl
nghiant2710: @petrosagg It comes from this thread https://gitter.im/resin-io/chat?at=57323356a351d8310952c20b. An user wanted to install some rpi specific packages but they are not available from this mirror http://archive.raspbian.org/raspbian jessie main contrib non-free rpi firmware. So I think we should add it to our raspberrypi base images

debian/generate-dockerfile.sh
nghiant2710: @petrosagg we need this change since `armel` is marked as `armv5e` in device types.

device-base/Dockerfile.armv7hf.rpi.tpl
nghiant2710: updated!

debian/generate-dockerfile.sh
petrosagg: This will probably affect the builder. Did you try pushing this to an ARMv5 application on resin?

nghiant2710: Yup, our builders expect `armv5e` not `armel`. Posted a thread about this change in FD

PR #131, opened 2016-06-07
Add node v6.2.1 and node v4.4.5

nghiant2710:


node/generate-stackbrew-library.sh
```diff
@@ -58,7 +58,7 @@ function generate_library(){
 
 declare -A aliases
 aliases=(
-	[6.2.0]='6 latest'
+	[6.2.1]='6 latest'
```
petrosagg: Shouldn't we also have aliases for 0, 0.10, 0.12, 4 and 5?

petrosagg:
SGTM

node/generate-stackbrew-library.sh
nghiant2710: updated!

Issue #137, opened 2016-06-29
/etc/hostname is being appended to on container restart


lorenzo-stoakes:
Currently, if a container is allowed to restart enough times, eventually an error is output indicating that the hostname is too long - `hostname: name too long`.

Experimenting with a Dockerfile consisting of:

```Dockerfile
FROM resin/raspberrypi3-debian

CMD hostname; echo "$HOSTNAME"; cat /etc/hostname
```

Gives output as shown below, which clearly indicates that `/etc/hostname` is having `-30fbcd` (the first 6 digits of the device UUID) being appended to on each restart.

```
29.06.16 12:33:13 [+0100] Started application 'registry.resin.io/test/c6be287555ad93a8f0cde5d6daf986cf53c8d1dc'
29.06.16 12:33:13 [+0100] raspberrypi3-30fbcd
29.06.16 12:33:13 [+0100] raspberrypi3-30fbcd
29.06.16 12:33:13 [+0100] raspberrypi3-30fbcd
29.06.16 12:33:13 [+0100]
29.06.16 12:33:14 [+0100] Application exited 'registry.resin.io/test/c6be287555ad93a8f0cde5d6daf986cf53c8d1dc'
29.06.16 12:33:14 [+0100] Restarting application 'registry.resin.io/test/c6be287555ad93a8f0cde5d6daf986cf53c8d1dc'
29.06.16 12:33:16 [+0100] raspberrypi3-30fbcd-30fbcd
29.06.16 12:33:16 [+0100] raspberrypi3-30fbcd-30fbcd
29.06.16 12:33:16 [+0100] raspberrypi3-30fbcd-30fbcd
29.06.16 12:33:16 [+0100]
29.06.16 12:33:16 [+0100] Application exited 'registry.resin.io/test/c6be287555ad93a8f0cde5d6daf986cf53c8d1dc'
29.06.16 12:33:16 [+0100] Restarting application 'registry.resin.io/test/c6be287555ad93a8f0cde5d6daf986cf53c8d1dc'
29.06.16 12:33:18 [+0100] raspberrypi3-30fbcd-30fbcd-30fbcd
29.06.16 12:33:18 [+0100] raspberrypi3-30fbcd-30fbcd-30fbcd
29.06.16 12:33:18 [+0100] raspberrypi3-30fbcd-30fbcd-30fbcd
29.06.16 12:33:18 [+0100]
[...]

PR #138, opened 2016-07-01
Some bugs fixes for base images

nghiant2710:


lorenzo-stoakes:
Confirmed fixes #137.

node/generate-stackbrew-library.sh
```diff
@@ -61,8 +61,7 @@ aliases=(
 	[6.2.2]='6 latest'
 	[5.11.1]='5'
 	[4.4.5]='4'
-	[0.12.14]='0.12 0'
-	[0.10.45]='0.10'
+	[0.12.14]='0'
```
lorenzo-stoakes: I'm not 100% clear on what this is used for, but is it necessary to specifically alias 0.12.14 to '0'? As 0.12.15 is the most recent version, would we specify that anywhere?

nghiant2710: We've not supported that version yet. 0.12.14 is the current latest 0.12.x

lorenzo-stoakes: Ok, in that case why do we alias it at all @nghiant2710 ?

nghiant2710: To make a tag with major version only that points to latest version.

lorenzo-stoakes: Ah ok I see

lorenzo-stoakes:
LGTM

lorenzo-stoakes:
(Am assuming you've tested the other components I couldn't test @nghiant2710 )

petrosagg:
SGTM

PR #143, opened 2016-07-05
Retire Systemd base images

nghiant2710:


debian/Dockerfile.systemd.partial
```diff
@@ -0,0 +1,24 @@
+# Install Systemd
```
petrosagg: Why don't we use a `Dockerfile.tpl` and `Dockerfile.nosystemd.tpl` just like it was in the systemd directory?

nghiant2710: To avoid redundancy since `Dockerfile.tpl` and `Dockerfile.nosystemd.tpl` will share some parts. Start scripting this way make it easier for me when migrating to doxx

petrosagg:
SGTM

PR #145, opened 2016-07-08
Add missing source.list entry for raspberrypi base images

nghiant2710:
Minimal raspbian doesn't include this entry but full version has it so we should have it in our base images


petrosagg:
SGTM

PR #146, opened 2016-07-13
Fix #136, mount /sys/fs/cgroup volume to fix systemd cgroup message

### LINKED ISSUE #136: Error message when systemd enabled
the following message "Failed to install release agent, ignoring: No such file or directory" appears after systemd enabled

nghiant2710:


petrosagg:
to avoid this fix getting lost, let's merge #143 and then put this on top. other than that LGTM

petrosagg:
SGTM

PR #151, opened 2016-07-20
Drop support for armv7hf sid base images

nghiant2710:


nghiant2710:
@petrosagg please review this PR

petrosagg:
@nghiant2710 this has conflicts. Why are we dropping sid?

nghiant2710:
@petrosagg I will resolve it later but I think we can drop sid since it's quite old, not sure if any users're using it. Sometimes, its repository mirrors are not reachable and our build jobs fail for it

petrosagg:
sid is not old at all. It is the unstable version of Debian and has all the latest packages

nghiant2710:
ok, I will close this PR

PR #152, opened 2016-07-21
Add raspberrypi-debian base image which is based on Debian armel

nghiant2710:


nghiant2710:
@petrosagg please review this PR

device-base/generate-dockerfile.sh
```diff
@@ -138,39 +140,37 @@ for device in $devices; do
 
 	# Debian.
 	debian_dockerfilePath="$device/debian"
-	if [ $device != "raspberrypi" ]; then
```
Page-: I love that it lets us get rid of this if :)

petrosagg:
@nghiant2710 can you alias it to raspbian instead? I give up purity

nghiant2710:
Close since this PR will be moved to resin-rpi-raspbian repo

PR #154, opened 2016-07-22
Various bug fixes for base images

nghiant2710:


python/generate-dockerfile.sh
```diff
@@ -38,6 +39,7 @@ append_setup_script() {
 # $1: base version
 # $2: path
 # $3: variants
+# S4: distro
```
Page-: This should be `$4`?

python/generate-dockerfile.sh
```diff
@@ -48,9 +50,10 @@ set_pythonpath() {
 		fi
 
 		if [ $1 != "2.7" ]; then
-			echo "ENV PYTHONPATH $PYTHON3_PATH" >> $tmp_path/Dockerfile
+			PPATH="PYTHON3_PATH_$4"
+			echo "ENV PYTHONPATH $(eval echo \$$PPATH):\$PYTHONPATH" >> $tmp_path/Dockerfile
```
Page-: Can't this just be `echo "ENV PYTHONPATH $PPATH:\$PYTHONPATH" >> $tmp_path/Dockerfile`?

python/generate-dockerfile.sh
nghiant2710: Updated!

python/generate-dockerfile.sh
nghiant2710: This must be the value of string variable (value of PPATH) not just PPATH

Page-: Ah, I get it, in that case you can use `echo "ENV PYTHONPATH ${!PPATH}:\$PYTHONPATH" >> $tmp_path/Dockerfile`

nghiant2710: Updated!

Page-:
SGTM

PR #164, opened 2016-08-08
Fix issue with wrong python version in fedora onbuild python base image.

nghiant2710:
Since we only have major version tag so onbuild images need to be built on top of major version tag instead of full version.


Page-:
SGTM

PR #165, opened 2016-08-09
[WIP] Switch to doxx

nghiant2710:


tools/lib/node.coffee
```diff
@@ -0,0 +1,62 @@
...
+RESIN_URL_PREFIX = 'http://resin-packages.s3.amazonaws.com/node/'
+NODEJS_URL_PREFIX = 'http://nodejs.org/dist/'
+
+UNSUPPORTED = 'UNSUPPORTED'
+
+ALPINE_ARCH =
+	'amd64': 'alpine-amd64'
+	'i386': 'alpine-i386'
+	'armv6hf': 'alpine-armhf'
+	'armv7hf': 'alpine-armhf'
```
petrosagg: @emirotin can this somehow be reflected in the dicts and not be js code?

tools/lib/utils.coffee
```diff
@@ -0,0 +1,61 @@
...
+
+resolveArch = (arch, distro) ->
+	if distro is 'alpine'
+		return ALPINE_ARCH[arch] or UNSUPPORTED
+	if distro is 'fedora'
+		return FEDORA_ARCH[arch] or UNSUPPORTED
+	return arch
+
+exports.getChecksum = (filename) ->
+	result = grep(filename, CHECKSUM_FILE_PATH)
```
petrosagg: @nghiant2710 this would be a lot better if the checksum file was JSON and it properly parsed it instead of grepping

tools/lib/node.coffee
emirotin: It would fit nicely if we had a dict for `arch`. You would just define a field, like `alpine_arch`, and later access it like `$arch.alpine_arch`.
But we don't have such dict.
We could put it in the device types dict, but then we'll have a bunch of duplicates for the devices that share the same arch.

We can improve `doxx` to have a new feature with dicts `FK`s, where the device type dict will reference the `arch` dict, which won't be used directly for templates expansion.
Like the `arch` dict would keep the `arch` details, like alpine mapping, and maybe some debian fixups we have below. And the device type dict will reference it like this: `$arch: "$arch.amd64"`. This reference will be expanded by `metalsmith-dynamic` and replaced with an actual object from the `$arch` dict.

emirotin: So, it should be working now. [https://github.com/resin-io-playground/metalsmith-dynamic#glossary]

Update to `metalsmith-dynamic@^0.4.0`
Create a new dictionary called `arch.json`. As the very minimum it should contain an array with objects `{ "id": "armvh7" }`, or the like.

Now in the device types dictionary switch from string arch refs to the new FKs:
from `"architecture": "armvh7"` to `"$architecture": "$arch.armvh7"`

Obviously in the source files you have to switch from `$device_type.architecture` to `$device_type.$architecture.id`.

Now you can add arbitrary props to the arch dictionary and use them in the HB helpers.

emirotin:
The PR is super-big again which makes it impossible to review, GH hides most of the files.
I suggest that we gitignore the generated files for now and keep the PR small. Then when it looks OK and is approved you can unignore them, add, commit, push and merge.

tools/lib/node.coffee
nghiant2710: Cool, great work @emirotin. I will update the PR

nghiant2710:
@emirotin, The PR is separated into 2 commits, the first one is source changes and the second one is autogenerated files. You can check the source changes only and ignore the second commit.
I don't want to remove autogenerated files since we also need to verify Dockerfile content

emirotin:
OK thanks, will check by commit

nghiant2710:
@petrosagg @emirotin I've fixed your comments and added commits for other base images. It would be great if I have your review before migrating python and golang base images since they basically will be the same as node

nghiant2710:
@petrosagg, ping

tools/README.md
```diff
@@ -0,0 +1,37 @@
+Dynamic Dockerfile demo
```
emirotin: The readme needs some updates

tools/build/build.coffee
```diff
@@ -0,0 +1,47 @@
...
+HbHelper.registerConsolidate(consolidate, importName: 'include')
+require('./hb-helpers')(HbHelper.Handlebars)
+
+metalsmith = Metalsmith(rootDir)
+.source('src')
+.destination(utils.destDir)
+.use(defaultPartialsSearch())
+.use(dynamic({
+  dictionaries: path.join(rootDir, 'dicts')
+  populateFields: [ '$partials_search' ]
```
emirotin: I know you've copied it from my example, but I'm not sure this line is actually needed. Can you check please?

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+path = require('path')
+fse = require('fs-extra')
+semver = require('semver')
+
+utils = require('../lib/utils')
+pythonHelper = require('../lib/python')
+
+
+removeFile = (obj, file, files) ->
+	if obj and obj is utils.UNSUPPORTED
```
emirotin: no need to check if `obj` is true-ish, can simply be: `if obj is utils.UNSUPPORTED`
`obj` is also not the best name here, should be like `isSupported` or something like that.

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+		delete files[file]
+
+copyFile = (src, dest) ->
+	try
+		fse.copySync(src, dest)
+	catch err
```
emirotin: this will log the error and swallow it, is it what we want?

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+		console.error(err)
+
+copyDebianFile = (obj, destDir) ->
```
emirotin: this may be the best solution at hand, but to me it feels really weird that we do some fs manipulations from the MS plugin. of course, you can do that, but it breaks the abstraction. MS plugins are supposed to manipulate the abstract files tree
instead of this one can use the existing mechanisms: put these files in the tree and annotate them properly to be orchestrated by MS. Like take the file `./supporting_files/debian.01_buildconfig`, place it like `src/debian.01_buildconfig` and add this front-matter

```
---
dynamic:
  variables: [ $debian_arch, $debian_suite ]
  ref: debian/$debian_arch/$debian_suite/01_buildconfig
  skip_ext: true
---
```

Same for other files. For files that are for example wheezy-only it can be

```
---
dynamic:
  variables: [ $debian_arch ]
  ref: debian/$debian_arch/wheezy/entry.sh
---
```

I see how there's a problem with "use this entry.sh for all _except of_ wheezy". This can be solved by having a placeholder file that imports a single partial `{{ import "debian-entry" }}`. That entry will have a `_default.part`, and then a single overload for `wheezy.part`.

Finally for files that need to be present for every suite except of wheezy (and be missing completely for wheezy) you can use the `UNSUPPORTED` trick placing this keyword into the partial itself (and then killing the files that contain this word in one final pass).

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+
+	# Systemd
+	if obj.$debian_suite.id is 'wheezy'
+		copyFile('./supporting_files/debian.entry-nosystemd.sh', "#{destDir}/entry.sh")
+	else
+		copyFile('./supporting_files/debian.entry.sh', "#{destDir}/entry.sh")
+		copyFile('./supporting_files/debian.launch.service', "#{destDir}/launch.service")
+
+	# QEMU
+	if obj.$debian_arch.id is 'aarch64'
```
emirotin: The approach suggested above doesn't work for binaries though, but I'd arrange this specific copying in the shell script after the entire tree is built.

tools/dicts/alpine_suite.json
```diff
@@ -0,0 +1,14 @@
...
+    "id": "3.4"
+  },
+  {
+    "id": "edge"
+  },
+  {
+    "id": "3.3"
+  },
+  {
+  	"id": "3.2"
```
emirotin: different indent (spaces/tabs)

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+		copyFile('./supporting_files/python.setup.sh', "#{destDir}/setup.sh")
+
+checkWheezyNode = (obj, file, files) ->
+	type = file.split('/')[0]
+	if type is 'node'
+		if semver.eq(obj.$node_version.id, '6.3.1')
+			# Only generate dockerfile for debian wheezy node v6.3.1
+			obj.$isSupported = utils.UNSUPPORTED
+			if obj.$distro.id is 'debian' and obj.$language_variant.id is 'old_suite'
+				obj.$isSupported = utils.SUPPORTED
```
emirotin: The `SUPPORTED` value is never used (never checked and seems to be only needed to erase the `unsupported` flag. I'd drop its usage. Just do:

```
if semver.eq(obj.$node_version.id, '6.3.1')
  if not (obj.$distro.id is 'debian' and obj.$language_variant.id is 'old_suite')
    obj.$isSupported = utils.UNSUPPORTED
```

and drop it in other places. The `$isSupported` name is a bit confusing then, though (it already is in the next line anyway). So it'd better be `$isUnsupported` IMO.

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+
+
+exports.expandProps = walkFiles (file, files) ->
+	obj = files[file]
+	propsToExpand = obj.expand_props
+	return if not propsToExpand
+	for prop in propsToExpand
+		obj[prop] = HbHelper.render(obj[prop], obj)
+	# expand special variable for each image type
+	type = file.split('/')[0]
```
emirotin: I'd move this to a separate plugin, it's currently mixing concerns.
I'd also better make it explicit. Set a variable in the files: `python: true`, and check it instead of relying on the path structure.

tools/build/ms-plugins.coffee
```diff
@@ -0,0 +1,113 @@
...
+copyGolangFile = (obj, destDir) ->
+	if obj.$language_variant.id != 'onbuild'
+		copyFile('./supporting_files/golang.go-wrapper', "#{destDir}/go-wrapper")
+
+copyPythonFile = (obj, destDir) ->
+	if obj.$device.id is 'edison' and obj.$distro.id is 'debian' and obj.$language_variant.id != 'onbuild'
+		copyFile('./supporting_files/python.setup.sh', "#{destDir}/setup.sh")
+
+checkWheezyNode = (obj, file, files) ->
+	type = file.split('/')[0]
```
emirotin: I'd probably move it to a HB helper that will be populated: `$isUnsupported: '{{ isUnsupported }}'`, and mark this field to be populated.

tools/README.md
nghiant2710: Updated the title, I will review the README content again before merging.

tools/build/build.coffee
nghiant2710: it's needed indeed, without it MS will use `_default` partial only

tools/build/ms-plugins.coffee
nghiant2710: updated

tools/build/ms-plugins.coffee
nghiant2710: yeah, I don't want an error can break the whole build, just log and we can fix later.

tools/dicts/alpine_suite.json
nghiant2710: updated

tools/build/ms-plugins.coffee
nghiant2710: updated

tools/build/ms-plugins.coffee
nghiant2710: I added `type` variable to store the base image type

tools/build/ms-plugins.coffee
nghiant2710: updated

nghiant2710:
@emirotin can you check this PR again?

PR #178, opened 2016-08-29
Fix #177, Set v6.3.1 as the latest version for debian wheezy node base images and add v6.5.0

### LINKED ISSUE #177: Node 6.4.0 doesn't work on debian wheezy base images
official v6.4.0 armv7l node binary requires `GLIBCXX_3.4.18` which is not available on wheezy images. While gcc/g++ 4.8 or newer is required to compile v6.4.0 so we can't compile it from wheezy images too. Therefore we need to rollback to v6.3 for debian wheezy base images.

nghiant2710:


petrosagg:
SGTM

PR #183, opened 2016-09-15
set QEMU_CPU so uname will report correct information

nghiant2710:


debian/entry.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+function remove_buildtime_env_var()
+{
+	unset QEMU_CPU
```
petrosagg: why unset the variable? We want it to persist

nghiant2710: @petrosagg, I think on device QEMU will no longer emulate and this command will report correct information without this env var

petrosagg: you're right. I thought this is build-time

petrosagg:
SGTM

PR #184, opened 2016-09-20
Add node v6.6.0

nghiant2710:


node/generate-stackbrew-library.sh
```diff
@@ -61,7 +61,7 @@ function generate_library(){
 
 declare -A aliases
 aliases=(
-	[6.5.0]='6 latest'
+	[6.6.0]='6 latest'
 	[6.3.1]='6 latest'
```
petrosagg: the alias for `6 latest` is defined twice

nghiant2710: the second alias is for debian wheezy only since later node versions don't work on wheezy anymore

PR #186, opened 2016-09-23
Fix issue with entry script in debian systemd base images

nghiant2710:
The recent change of the entry script is a breaking change (can not start interactive shell unless overriding current entrypoint since CMD will be started in a subshell) We should keep the old behavior since our base images are widely used by the community, also it is easier to run it locally.

### DIFF
--- a/debian/aarch64/jessie/entry.sh
+++ b/debian/aarch64/jessie/entry.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 
+function start_udev()
+{
+	udevd & 
+	udevadm trigger &> /dev/null
+}
+
 function remove_buildtime_env_var()
 {
 	unset QEMU_CPU
@@ -65,16 +71,18 @@ function init_non_systemd()
 	# trap the stop signal then send SIGTERM to user processes
 	trap signal_handler SIGRTMIN+3 SIGTERM
 
-	udevd & 
-	udevadm trigger &> /dev/null
-	
 	CMD=$(which "$1")
 	# echo error message, when executable file doesn't exist.
 	if [ $? == '0' ]; then
 		shift
+		if [ ! -z "$RESIN" ] && [ ! -z "$RESIN_DEVICE_UUID" ]; then
+			# run this on resin device only
+			start_udev
+		fi
 		"$CMD" "$@" &
 		pid=$!
 		wait $pid
+		fg &> /dev/null
 	else
 		echo "Command not found: $1"
 		exit 1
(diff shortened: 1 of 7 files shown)


debian/entry.sh
```diff
@@ -72,9 +72,14 @@ function init_non_systemd()
 	# echo error message, when executable file doesn't exist.
 	if [ $? == '0' ]; then
 		shift
-		"$CMD" "$@" &
-		pid=$!
-		wait $pid
+		if [ ! -z "$RESIN" ] && [ ! -z "$RESIN_DEVICE_UUID" ]; then
```
petrosagg: This means that an interactive command will always start a udev process. Can you isolate just the command running part in its own function and call that when we're not in a resin context?

nghiant2710: Updated!

debian/aarch64/jessie/entry.sh
```diff
@@ -65,16 +71,20 @@ function init_non_systemd()
...
-		wait $pid
+		if [ ! -z "$RESIN" ] && [ ! -z "$RESIN_DEVICE_UUID" ]; then
+			# run this on resin device only
+			start_udev
+			
+			"$CMD" "$@" &
+			pid=$!
+			wait $pid
+		else
+			exec "$CMD" "$@"
```
petrosagg: Unfortunately we can't use `exec` without breaking `docker stop`. Since we declared that we need signal RTMIN+3 the user process won't know how to handle it. So we need to use `wait $pid` on this case too. Only udev should be kept in the conditional

nghiant2710: @petrosagg we can't solve the interactive shell issue, if we do that since it's started as subshell. So I think it's better to note down RTMIN+3 stop signal in README so users will know and handle it manually.

petrosagg: You can. just switch `wait $pid` with `bg`

nghiant2710: Updated!

debian/entry.sh
```diff
@@ -65,16 +71,18 @@ function init_non_systemd()
...
 	if [ $? == '0' ]; then
 		shift
+		if [ ! -z "$RESIN" ] && [ ! -z "$RESIN_DEVICE_UUID" ]; then
+			# run this on resin device only
+			start_udev
+		fi
 		"$CMD" "$@" &
 		pid=$!
 		wait $pid
+		fg &> /dev/null
```
petrosagg: Why are you redirecting output to /dev/null? You want the stdout/stderr streams of this process to be attached to whoever started the container. Being an interactive shell or the docker log management

nghiant2710: it only redirects output of fg command, which will confused users, to /dev/null, the output from CMD command still attaches to the sheel or log management

petrosagg: Nice. Didn't know that

PR #199, opened 2016-10-23
Add support for new board: intel-quark

nghiant2710:


nghiant2710:
Connects to https://github.com/resin-io/docs/issues/344 - golang 1.5.x won't work even on MMX because the compilers emit invalid instructions should be mentioned in our docs

PR #200, opened 2016-10-25
Add support for new board Artik710

nghiant2710:


device-base/Dockerfile.artik.fedora.tpl
```diff
@@ -4,6 +4,8 @@ LABEL io.resin.device-type="#{DEV_TYPE}"
 
 COPY artik.repo /etc/yum.repos.d/
 COPY RPM-GPG-KEY-artik-#{SUITE}-armhfp /etc/pki/rpm-gpg/
+COPY zigbee-daemon.service zigbee-pre.service /lib/systemd/system/
+COPY zigbee_userspace.tar.gz /
```
petrosagg: Is this not available from an upstream location somewhere? Do we have to vendor it in our repo?

PR #212, opened 2016-12-01
Update debian i386 node binary with no-MMX patch

nghiant2710:


nghiant2710:
connects to https://github.com/resin-io/node-arm/pull/35

PR #215, opened 2016-12-15
Update pip to v9.0.1

nghiant2710:


Page-:
Is this a breaking change to our python images (I see the major version bump)? cc @shaunmulligan

nghiant2710:
@Page-, pip version bump won't be a breaking change. pip v9.x won't break anything in our base images

shaunmulligan:
@Page- looking at the release notes, it seems it should be okay: https://pip.pypa.io/en/stable/news/

the main one I see is this:
>BACKWARD INCOMPATIBLE Remove the attempted autodetection of requirement names from URLs, URLs must include a name via #egg=.

PR #219, opened 2017-01-03
Add Alpine Linux 3.5 and aarch64 alpine linux base images

nghiant2710:


nghiant2710:
ping @imrehg, @shaunmulligan, we will have aarch64 alpine linux base images soon

PR #225, opened 2017-01-13
Fix unmet dependencies when installing openjdk.

nghiant2710:


openjdk/Dockerfile.tpl
```diff
@@ -36,7 +36,7 @@ ENV JAVA_DEBIAN_VERSION #{DEBIAN_VERSION}
 
 RUN set -x \
 	&& apt-get update \
-	&& apt-get install -y \
+	&& apt-get install -y #{DEBIAN_PRIOTIZED_SOURCE} \
```
petrosagg: typo priotized -> prioritized

nghiant2710: @petrosagg Updated!

cskiraly:
I was testing this PR on `openjdk/raspberrypi2/debian/8-jdk/Dockerfile`, which was not building before. 
With the fix the image builds (even in emulation). However, when used as a base image, the fix creates problems with apt-get, such as:
```
apt-get install libglib2.0-dev
[...]
The following packages have unmet dependencies:
 libglib2.0-dev : Depends: libglib2.0-0 (= 2.42.1-1+b1) but 2.48.0-1~bpo8+1 is to be installed
                  Depends: libglib2.0-bin (= 2.42.1-1+b1) but it is not going to be installed
```

`apt-get install -t jessie-backports libglib2.0-dev` works, but as a base image, this looks cumbersome. Any ideas how to fix this pinning such that apt-get works afterwards as usual?

(As an side note, I was also trying to build the base image without this PR, and without specifying the java version string in apt-get install. It currently builds fine.)

nghiant2710:
@cskiraly, I see you problem but our base images base on the official base images from Docker (https://github.com/docker-library/openjdk) so we don't want to change the format of the Dockerfile. I think we can pin the jessie-backports repository with smaller priority so that apt will not select package from that repository if we don't specify it. I will update this PR to fix this

PR #227, opened 2017-01-16
Fix issue with SIGTERM is not properly handled if INITSYSTEM not enabled.

nghiant2710:
Connects to https://github.com/resin-io-library/resin-rpi-raspbian/pull/48

### DIFF
--- a/debian/entry.sh
+++ b/debian/entry.sh
@@ -83,7 +83,7 @@ function init_non_systemd()
 		fi
 		"$CMD" "$@" &
 		pid=$!
-		fg > /dev/null
+		wait "$pid"
 	else
 		echo "Command not found: $1"
 		exit 1
--- a/debian/aarch64/jessie/entry.sh
+++ b/debian/aarch64/jessie/entry.sh
@@ -83,7 +83,7 @@ function init_non_systemd()
 		fi
 		"$CMD" "$@" &
 		pid=$!
-		fg > /dev/null
+		wait "$pid"
 	else
 		echo "Command not found: $1"
 		exit 1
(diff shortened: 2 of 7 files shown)


debian/entry.sh
```diff
@@ -83,6 +83,7 @@ function init_non_systemd()
 		fi
 		"$CMD" "$@" &
 		pid=$!
+		wait "$pid"
```
petrosagg: If you wait here for `$pid`, then what does the `fg` immediately afterwards do? Can you explain why the signal does not properly handled? (and preferably add this info in the commit message in the end).

nghiant2710: @petrosagg yeah, `fg` should be removed. The explanation was added in the original commit https://github.com/resin-io-library/resin-rpi-raspbian/pull/48, I also added it here

petrosagg:
I realised we have a much bigger problem here. The behaviour we observe when we use `fg` instead of `wait` is actually the default behaviour. For example, the following script will never respond to SIGTERM:

```bash
#!/bin/bash

set -m

function signal_handler() {
	echo "signal handler"
}
trap signal_handler SIGTERM
sleep 1000
```

The problem though is that while we can use `wait` to be able to respond to signals that arrive to the container, we're currently ignoring any child process that is re-parented to PID 1 (i.e entry.sh) because its parent crashed. This breaks expectations because Docker users expect orphaned processes to be reparented to their own PID 1.

I think it would be a good idea to integrate [tini](https://github.com/krallin/tini) in our images until we upgrade docker on devices to 1.13 which has this feature built-in.

nghiant2710:
@petrosagg indeed, we need tini in our base images, I will integrate it in our base images when I'm back next week. At the moment, can we merge this PR as a temporary fix?

PR #237, opened 2017-02-17
Integrate Tini as the default init for our base images

nghiant2710:
https://github.com/krallin/tini


shaunmulligan:
@nghiant2710 how will this affect users? does this mean tini will start systemd when INITSYSTEM=on?

alpine/generate-dockerfile.sh
```diff
@@ -56,10 +56,19 @@ for repo in $repos; do
 		dockerfilePath=$repo/$suite
 		mkdir -p $dockerfilePath
 
+		if version_le $suite "3.4"; then
```
petrosagg: why are we not using tini in this case?

debian/generate-dockerfile.sh
```diff
@@ -93,6 +98,21 @@ for arch in $archs; do
 			cat Dockerfile.systemd.partial >> $dockerfilePath/Dockerfile
 			cp entry.sh launch.service $dockerfilePath/
 		fi
+
+		# Tini
+		# TODO: armel support.
```
petrosagg: Can we add armel support?

alpine/generate-dockerfile.sh
nghiant2710: Because tini is not available in Alpine Linux 3.4, however I removed this since we dropped support for Alpine Linux 3.4

debian/generate-dockerfile.sh
nghiant2710: Added!

nghiant2710:
@petrosagg I've updated the PR, can you please review it again?

alpine/generate-dockerfile.sh
```diff
@@ -58,8 +58,9 @@ for repo in $repos; do
 
 		sed -e s~#{FROM}~"$baseImage:$suite"~g \
 			-e s~#{LABEL}~"$label"~g \
-			-e s~#{QEMU}~"$qemu"~g Dockerfile.tpl > $dockerfilePath/Dockerfile
-		cp entry.sh resin rc.conf resin-xbuild $dockerfilePath/
+			-e s~#{QEMU}~"$qemu"~g \
+			-e s~#{PACKAGES}~"$packages"~g Dockerfile.tpl > $dockerfilePath/Dockerfile
```
petrosagg: `PACKAGES` are always an empty string. Why do we add this?

nghiant2710: Tini is available from Alpine 3.4 so we only install it for Alpine 3.4 and higher but we dropped support for old versions so It can be removed. I removed it.

PR #238, opened 2017-02-17
Fix #250, add `fg`command to fix the broken interactive shell when runs container locally.

### LINKED ISSUE #228: INITSYSTEM should also accept 1 or 0 as well as on and off
relates to https://github.com/resin-io/hq/issues/364

### LINKED ISSUE #250: Unable to launch interactive shell when running the container locally
With the current implementation https://github.com/resin-io-library/base-images/blob/master/debian/entry.sh#L84-L86, we will never be able to start an interactive shell when running the container locally, the reason is CMD will be started in background so command like `docker run -it resin/raspberrypi3-debian:jessie /bin/bash` won't work as expected. This is really bad since our base images are widely used by Docker community.

Adding `fg` command after having run `wait $pid` is the solution here but there is another issue since ` `fg` always returns exit code 1 if there is no background process to attach to so we need a way to pass the correct exit code in that case.

There were few PRs to fix this issue:
https://github.com/resin-io-library/base-images/pull/186 (https://github.com/resin-io-library/base-images/commit/b18bd44badeccae9cf7b549c96b6fcb2f861bfdc)
https://github.com/resin-io-library/base-images/pull/191 (https://github.com/resin-io-library/base-images/commit/b789b41f43b5d92d069e2ec0878f0b130c90947a)
https://github.com/resin-io-library/base-images/pull/227 (https://github.com/resin-io-library/base-images/commit/4be4d71862db662c603aa090ed2d8f9ac69d5c96)
but none of them completely solves this.

This PR https://github.com/resin-io-library/base-images/pull/238 will address all the problems that relate to this issue.

nghiant2710:


alpine/entry.sh
```diff
@@ -1,11 +1,16 @@
...
 
+# On ResinOS 2.x devices, the hostname is set by the hostOS.
+# For backward compatibility, we only update the hostname for ResinOS 1.x devices.
 function update_hostname()
 {
-	HOSTNAME="$HOSTNAME-${RESIN_DEVICE_UUID:0:7}"
-	echo $HOSTNAME > /etc/hostname
-	echo "127.0.1.1 $HOSTNAME" >> /etc/hosts
-	hostname "$HOSTNAME"
+	if [ ${RESIN_DEVICE_UUID:0:7} != $(hostname -s | cut -c 1-7) ]; then
```
Page-: Can this be:
```bash
if [ ${RESIN_DEVICE_UUID:0:7} != ${HOSTNAME:0:7} ]; then
```
in order to make it a bit easier to see that the same slice is being compared

alpine/entry.sh
```diff
@@ -61,6 +61,23 @@ function init_non_openrc()
 	fi
 }
 
+INITSYSTEM=$(echo "$INITSYSTEM" | awk '{print tolower($0)}')
+
+case "$INITSYSTEM" in
+	'1')
```
Page-: Is this able to be
```bash
	'1' | 'true')
		INITSYSTEM='on'
```
?

nghiant2710: Updated!

alpine/entry.sh
nghiant2710: Updated!

debian/entry.sh
```diff
@@ -90,6 +90,8 @@ function init_non_systemd()
 		"$CMD" "$@" &
 		pid=$!
 		wait "$pid"
+		exit_code=$?
+		fg > /dev/null || exit "$exit_code"
```
petrosagg: `fg` was not deleted in the commit referenced in your commit message. Can you open an issue in this repo that clearly states the problem you're trying to fix here, including a minimal testcase? I don't understand what's going on with the exit codes and what we're trying to fix

nghiant2710: @petrosagg https://github.com/resin-io-library/base-images/issues/250

petrosagg: this PR doesn't have a "fixes #250" anywhere. Also, the issue doesn't explain why the issue happens, it just states that "interactive shell doesn't work". Can you explain the reason and what is going on with the script, the job, background, foreground etc?

nghiant2710: @petrosagg I've update the commit message and https://github.com/resin-io-library/base-images/issues/250 to explain more about this issue.

petrosagg: How does this fix the case where a user does `docker run -it ... /bin/bash`? From what I understand `/bin/bash` will be run in the background and then we'll block at `wait "$pid"`. And only after this happens will `fg` run.

nghiant2710: @petrosagg, when '/bin/bash' is started in the background that means it will immediately receive a SIGTSTP on start which puts the process on hold. Then `wait` command will catch this state change and pass on to `fg` which brings the shell back to foreground.

PR #247, opened 2017-03-10
Some updates for Node and Python base images

nghiant2710:


nghiant2710:
@Page- @petrosagg since this PR is crucial, I will merge it now. We can revert it later if you find anything wrong.

PR #256, opened 2017-03-28
Fix #216, Update the way hostname is set.

### LINKED ISSUE #216: Update the way the hostname is set
ResinOS 2.x changed the way the hostname is set. See https://github.com/resin-os/meta-resin/issues/509 . Discussed on Flowdock.

nghiant2710:


alpine/entry.sh
```diff
@@ -1,11 +1,17 @@
...
 function update_hostname()
 {
-	HOSTNAME="$HOSTNAME-${RESIN_DEVICE_UUID:0:7}"
-	echo $HOSTNAME > /etc/hostname
-	echo "127.0.1.1 $HOSTNAME" >> /etc/hosts
-	hostname "$HOSTNAME"
+	HOSTNAME=$(hostname)
+	if [ ${RESIN_DEVICE_UUID:0:7} != ${HOSTNAME:0:7} ]; then
+		# For 1.x Devices only.
+		HOSTNAME="$HOSTNAME-${RESIN_DEVICE_UUID:0:7}"
```
petrosagg: Can't we base the hostname only to variables that are set by the supervisor? For example `HOSTNAME="$RESIN_DEVICE_TYPE-${RESIN_DEVICE_UUID:0:7}"`

nghiant2710: Makes sense, @petrosagg. Updated!

shaunmulligan: @petrosagg and @nghiant2710 in that case what happens when an image isn't started by the supervisor?

nghiant2710: @shaunmulligan, The entryscript only calls this function if it's running on a resin device https://github.com/resin-io-library/base-images/blob/master/debian/entry.sh#L103-L107

shaunmulligan: @nghiant2710 okay cool, makes sense!

PR #260, opened 2017-04-12
Change the base image for all Debian images to slim variant.

nghiant2710:
it is a slimmer base (removing extra files, docs or manpages). This will save us 20-30 MB in comparison to current base.


petrosagg:
@nghiant2710 is there a spec for this? cc @shaunmulligan

nghiant2710:
Fixes https://github.com/resin-io-library/base-images/issues/265

PR #261, opened 2017-04-14
Remove unnecessary Raspbian repository.

nghiant2710:
We should keep the master Raspbian repository only for RPi family base
images.

### DIFF
--- a/device-base/Dockerfile.armv7hf.rpi.tpl
+++ b/device-base/Dockerfile.armv7hf.rpi.tpl
@@ -2,10 +2,8 @@ FROM #{FROM}
 
 LABEL io.resin.device-type="#{DEV_TYPE}"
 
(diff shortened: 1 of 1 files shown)


petrosagg:
@nghiant2710 why?

nghiant2710:
@petrosagg while I was reviewing the sources lists in the RPI family base images (to remove custom resin repository) found out that the `raspberrypi.org` repo and `raspbian.org` repo provide almost the same packages.
Actually `raspberrypi.org` has few packages outside `raspbian.org` and sometimes provide newer version than the other repo. Normally `apt` will prioritise the latest package and will install them from `raspberrypi.org`. I'm not sure this is the behavior we want so I think it's better to keep the Raspbian mirror only. We will mention manually adding the `raspberrypi.org` repository in our docs in case any users need a specific package from it.

nghiant2710:
we will keep both repositories so I will close this PR.

PR #264, opened 2017-04-16
Update the proper way to install pip and setuptools on Python base images

nghiant2710:


shaunmulligan:
what issue does this fix, if any? what changes/affects will this have for the users of python base images?

nghiant2710:
@shaunmulligan this won't affect users at all. It reduces the build time for each Python image and way easier to maintain. Building pip from source is not necessary and too complex.

shaunmulligan:
cool, good to know

nghiant2710:
Fixes https://github.com/resin-io-library/base-images/issues/266

nghiant2710:
`get-pip.py` was downloaded from `https://bootstrap.pypa.io/get-pip.py` and uploaded to our bucket. It's a recommended way to install pip and setuptools since we don't want to build them from source anymore.

python/Dockerfile.alpine.python3.slim.tpl
```diff
@@ -42,24 +40,14 @@ RUN set -x \
...
-	&& curl -SL "https://github.com/pypa/pip/archive/$PYTHON_PIP_VERSION.tar.gz" -o pip.tar.gz \
-	&& echo "$PYTHON_PIP_SHA256  pip.tar.gz" > pip.tar.gz.sha256sum \
-	&& sha256sum -c pip.tar.gz.sha256sum \
-	&& tar -xzC /usr/src/python/pip --strip-components=1 -f pip.tar.gz \
-	&& rm pip.tar.gz* \
-	&& cd /usr/src/python/pip \
-	&& python3 setup.py install \
-	&& cd .. \
+	&& if [ ! -e /usr/local/bin/pip3 ]; then : \
+		&& curl -SLO "http://resin-packages.s3.amazonaws.com/python/get-pip.py" \
```
petrosagg: Can we download the script from here so that we don't have to host it ourselves? https://raw.githubusercontent.com/pypa/get-pip/430ba37776ae2ad89f794c7a43b90dc23bac334c/get-pip.py

nghiant2710: @petrosagg Updated!

PR #276, opened 2017-05-08
Fix #273, Create empty man directory to fix missing manpage directory issue

### LINKED ISSUE #273: Errors installing OpenJDK due to unexistent man pages directory
```
[Build]    aspell-autobuildhash: processing: en [en_US-wo_accents-only].
[Build]    Processing triggers for ca-certificates (20141019+deb8u2) ...
[Build]    Updating certificates in /etc/ssl/certs... 0 added, 0 removed; done.
[Build]    Running hooks in /etc/ca-certificates/update.d....
[Build]    done.
[Build]    done.
[Build]    Processing triggers for menu (2.1.47) ...
[Build]    Errors were encountered while processing:
[Build]     openjdk-7-jre-headless:amd64
[Build]     openjdk-7-jre:amd64
[Build]     default-jre
[Build]     arduino
[Build]    E: Sub-process /usr/bin/dpkg returned an error code (1)
[Build]    Removing intermediate container f6f6f937a059
[Error] Build failed: The command '/bin/sh -c apt-get update && apt-get install -y --no-install-recommends xorg mono-complete libasound2-dev fluxbox dbus arduino g++ gcc usbutils make && rm -rf /var/lib/apt/lists/*' returned a non-zero code: 100
```

And the root cause is:

```
[Build] update-alternatives: error: error creating symbolic link `/usr/share/man/man1/rmid.1.gz.dpkg-tmp': No such file or directory
```

The `/usr/share/man/man1` directory doesn't exist, and simply doing `mkdir /usr/share/man/man1` before installing OpenJDK fixes the issue.

Maybe this is something we can do by default?

nghiant2710:


debian/Dockerfile.tpl
```diff
@@ -17,3 +17,5 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 COPY 01_nodoc /etc/dpkg/dpkg.cfg.d/
 COPY 01_buildconfig /etc/apt/apt.conf.d/
+
+RUN mkdir -p /usr/share/man/man1
```
petrosagg: Why is this missing in the first place?

nghiant2710: @petrosagg it is excluded in this configuration file https://github.com/resin-io-library/base-images/blob/master/debian/01_nodoc#L4

petrosagg: Isn't this a bug then? We should fix it upstream, not here

petrosagg: @nghiant2710 bump

nghiant2710: raised here https://github.com/tianon/docker-brew-debian/issues/65

PR #277, opened 2017-05-08
Fix #274, Add quotes to env var value in docker.env file

### LINKED ISSUE #274: Environment variables should be added to docker.env between quotes
Otherwise doing `source docker.env` will fail if there's variables with spaces.
Which happens with RESIN_HOST_OS_VERSION on resin.io devices.

nghiant2710:


debian/entry.sh
```diff
@@ -58,7 +58,7 @@ function init_systemd()
 {
 	GREEN='\033[0;32m'
 	echo -e "${GREEN}Systemd init system enabled."
-	env > /etc/docker.env
+	env | sed 's/"/\\"/g' | sed 's/^\([a-zA-Z0-9_]*\)=\(.*\)$/\1="\2"/g' > /etc/docker.env
```
petrosagg: Can we use shell escaping instead? I haven't tested if it works but if it does it will take care of all the edge cases.
```bash
for var in $(compgen -e); do
    printf '%q=%q\n' "$var" "${!var}"
done > /etc/docker.env
```

nghiant2710: Updated!

debian/amd64/jessie/entry.sh
```diff
@@ -58,7 +58,9 @@ function init_systemd()
 {
 	GREEN='\033[0;32m'
 	echo -e "${GREEN}Systemd init system enabled."
-	env > /etc/docker.env
+	for var in $(compgen -e); do
+    	printf '%q="%q"\n' "$var" "${!var}"
```
petrosagg: all oft these should be indented one tab level in. consider it approved once this is fixed

PR #288, opened 2017-06-04
Update QEMU to v2.9.0+resin1

nghiant2710:
resin-xbuild is built from this branch https://github.com/resin-io-projects/armv7hf-debian-qemu/tree/xbuild, we should move this branch to a new repo under `resin-io-library`


alpine/Dockerfile.tpl
```diff
@@ -3,6 +3,10 @@ FROM #{FROM}
 #{LABEL}
 
 #{QEMU}
+COPY resin-xbuild /usr/bin/
+RUN ln -s resin-xbuild /usr/bin/cross-build-start \
+	&& ln -s resin-xbuild /usr/bin/cross-build-end \
+	&& ln /bin/sh /bin/sh.real
```
petrosagg: You don't need to link this

debian/Dockerfile.tpl
```diff
@@ -7,6 +7,10 @@ ENV DEBIAN_FRONTEND noninteractive
 
 #{QEMU}
 #{QEMU_CPU}
+COPY resin-xbuild /usr/bin/
+RUN ln -s resin-xbuild /usr/bin/cross-build-start \
+	&& ln -s resin-xbuild /usr/bin/cross-build-end \
+	&& ln /bin/sh /bin/sh.real
```
petrosagg: same here

alpine/Dockerfile.tpl
nghiant2710: @petrosagg, Does `resin-xbuild` require `sh.real`? https://github.com/resin-io-projects/armv7hf-debian-qemu/blob/master/resin-xbuild.go#L12

petrosagg: yes, but it will create it if it doesn't exist so you don't need to do it here

nghiant2710: @petrosagg, updated!

debian/Dockerfile.tpl
nghiant2710: updated.

PR #293, opened 2017-06-15
Change the base of Resin base images to Docker official repositories.

nghiant2710:
List of officially supported archs https://github.com/docker-library/official-images#architectures-other-than-amd64.


imrehg:
A thing I'm not sure about is that `armhf` alpine images now coming from `arm32v6/alpine` which is for the Pi1 and Zero, there are no `arm32v7/alpine` images (yet?). This is in contrast with e.g. the Debian images which are from `arm32v7`.

nghiant2710:
@imrehg, I agree however there are not `arm32v7/alpine` at the moment and the `arm32v6/alpine` image works properly on armv7 devices (I tested on the boards I have: RPI2, RPI3 and have no issues) so I think we should use it now and change to `arm32v7` when they release.

PR #298, opened 2017-06-29
[WIP] Refactoring base images

nghiant2710:


nghiant2710:
@alexandrosm IMO, the blob entity you want is `stack_version`. Basically, I think `stack_version` (https://github.com/resin-io-library/base-images/pull/298/commits/28db079e25022f70265f6d0a745c90b6dd8b04f7#diff-553e99faf730145755967335765dce0fR124) and the original `blob` (https://github.com/resin-io-library/base-images/pull/298/commits/28db079e25022f70265f6d0a745c90b6dd8b04f7#diff-553e99faf730145755967335765dce0fR124) are the same and they can't represent a binary object. I think what I have now can be called blob_of_stack_supports_arch_is_compiled_for_libc.

If we want to find all "Python 2.7" blobs, just look for all blobs connect with `stack_version` 10, for example.

alexandrosm:
@nghiant2710 in this way we can't know if a blob is missing from a stack or not. We're trusting the dataset maker to do it right. But even if we do it right, if a stack has multiple blobs, we have no way of differentiating them. If the python stack is composed of the python executable and the pip executable (for instance), we don't know which is which unless we parse the url.

PR #311, opened 2017-07-18
Fix issue with checksum line format on Debian Wheezy base images.

nghiant2710:


petrosagg:
I don't understand what this PR is fixing. Where do we use `%` as a field separator?

nghiant2710:
@petrosagg, the issue here was `sha256sum` on wheezy is an old version and it only accepts checksum line with this format: `<checksum><space><space><filename>` (later versions accept both single or two spaces).
The checksum strings have 2 spaces https://github.com/resin-io-library/base-images/blob/master/debian/generate-dockerfile.sh#L13-L18 however they are trimmed so we need to set the field separator to something else to preserve two spaces.

debian/generate-dockerfile.sh
```diff
@@ -75,7 +75,9 @@ for arch in $archs; do
 
 	# Tini
 	tiniBinary="tini$TINI_VERSION.linux-$arch.tar.gz"
+	IFS='%'
 	tiniChecksum="TINI_$arch" && tiniChecksum=$(eval echo \$$tiniChecksum)
```
petrosagg: Setting the IFS is very hacky. You can instead do `tiniChecksum=$(eval echo \"\$$tiniChecksum\")`. Also I don't understand why you don't do `tiniChecksum=$(eval echo \"\$TINI_$arch\")` directly

nghiant2710: updated!

PR #312, opened 2017-07-18
Drop support for edge tag on Alpine Linux base images

nghiant2710:
Since the official Alpine docker repository no longer provides edge tag then we drop support for this tag.


petrosagg:
@shaunmulligan can you take a look here? Is it ok to drop support for this?

PR #314, opened 2017-07-20
Fix #313, Only generate node v0.10.x armel base images

### LINKED ISSUE #313: Node version greater than 0.10.x can't run on armv5 device.
Starting with the 3.18 release, the minimal V8 requirements will increase to ARMv6 + VFPv2 (https://github.com/v8/v8/blob/master/ChangeLog#L10506) which means it drops support for armv5. So only node v0.10.x can run on armv5 devices.

We've not found this before since all build process runs on emulation environment, for example: we can compile the armv5 binary for node v4.x and run it in `resin/armel-debian` image with QEMU but the node binary will crash if we specify the correct QEMU_CPU for armv5 (QEMU_CPU=arm1026). So we should drop all node versions except 0.10.x on armv5 base images.

nghiant2710:


petrosagg:
@nghiant2710 please add justification for this change

nghiant2710:
@petrosagg I've explained the issue here https://github.com/resin-io-library/base-images/issues/313

petrosagg:
Thanks. Can you annotate the PR or commits with the appropriate "Fixes: X" tag? In general try to keep things linked together as it helps during review and as future reference.

PR #315, opened 2017-07-22
Fix #299, add support for zulu variant openjdk base images

### LINKED ISSUE #299: Create ZuluJDK variant for Java images
On some platforms ZuluJDK has superior performance:

```
Native build performance on RPi3 + Samsung EVO+ 32GB SD:
  mvn package of recommenderandconfigurator
  - with OpenJDK:                                            10 minutes
  - with ZuluJDK in default mixed mode:            1 minutes
  - with ZuluJDK in interpreted mode (-Xint):  5.5 minutes

Runtime performance
  Spring Boot startup times for the recommenderandconfigurator service
  - with OpenJDK:                                            180 seconds
  - with ZuluJDK in default mixed mode:            13.5 seconds
  - with ZuluJDK in interpreted mode (-Xint):     66.5 seconds
```

nghiant2710:
Zulu source is not public so we only support zulu openjdk version 8 on Debian for armv7hf, armv6hf and amd64. Will try to get more versions supported later.


openjdk/generate-dockerfile.sh
```diff
@@ -50,7 +76,7 @@ for device in $devices; do
 			deviceArch='armhf'
 		;;
 		'raspberry-pi2')
-			deviceArch='armhf'
+			deviceArch='armhf'
```
petrosagg: trailing white space here

nghiant2710: Updated!

PR #335, opened 2017-09-05
Add debian backports repository for up-board base images

nghiant2710:
Fix https://github.com/resin-os/resin-up-board/issues/26


petrosagg:
This doesn't sound right. Why not use debian:stretch to get a newer stable mesa library?

PR #338, opened 2017-09-14
Fix issue with service is not properly cleaned up on Alpine Linux images

nghiant2710:
Connects to https://github.com/resin-io/hq/issues/987.
After checking user's log and Dockerfile, this issue happens when container starts and can only be reproduced by creating a process and echoing its PID to `/var/run/resinapp.pid` before starting OpenRC. 

When OpenRC starts resin service, it will  first check whether the process with PID in `/var/run/resinapp.pid` is running or not (if that file exists) and the error will be thrown in case it's running. It seems weird that the issue can happen right after the container starts so I think the possible fix here is to make sure that the pidfile and process don't exist.

### DIFF
--- a/alpine/aarch64/3.5/entry.sh
+++ b/alpine/aarch64/3.5/entry.sh
@@ -44,6 +44,11 @@ function init_openrc()
 	printf '%q ' "$@" >> /etc/resinApp.sh
 	chmod +x /etc/resinApp.sh
 
+	# Make sure there is no pidfile before starting.
+	if [ -f "/var/run/resinapp.pid" ]; then
+		rm -f /var/run/resinapp.pid &> /dev/null
+	fi
+
 	sed -i -e s~#{DIR}~"$(pwd)"~g /etc/init.d/resin
 
 	exec /sbin/init quiet
--- a/alpine/aarch64/3.6/entry.sh
+++ b/alpine/aarch64/3.6/entry.sh
@@ -44,6 +44,11 @@ function init_openrc()
 	printf '%q ' "$@" >> /etc/resinApp.sh
 	chmod +x /etc/resinApp.sh
 
+	# Make sure there is no pidfile before starting.
+	if [ -f "/var/run/resinapp.pid" ]; then
+		rm -f /var/run/resinapp.pid &> /dev/null
+	fi
+
 	sed -i -e s~#{DIR}~"$(pwd)"~g /etc/init.d/resin
 
 	exec /sbin/init quiet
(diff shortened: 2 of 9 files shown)


petrosagg:
This doesn't sound like a problem. Only the first process should get the signal

nghiant2710:
@petrosagg, the current implementation `of resinApp.sh (`#!/bin/bash\n exec $@`) will just replace the running shell (resinApp.sh) with whatever users put in the CMD (first process), however, OpenRC will only kill `resinApp.sh` on service restart or stop since it's in the service configuration file and so the first process will never receive the signal. This PR will fix the issue by trapping resinApp exit and send signal to the first process.

petrosagg:
@nghiant2710 I don't understand the problem. After `exec` `resinApp.sh` doesn't exist. OpenRC will send the signal to the exec'd process

nghiant2710:
@petrosagg OpenRC will only send the signal to the process with pid in pidfile (/var/run/resinapp.pid) which is resinApp.sh (https://github.com/resin-io-library/base-images/blob/master/alpine/resin#L7-L16). If it doesn't exit then it won't send the signal to the exec'd process.

petrosagg:
When you start `resinApp.sh`, it gets a PID. When it does `exec foobar`, then its process gets replaced with `foobar`, but foobar will have **the same** PID. If you observe that the PID is changing is because whatever you're exec'ing is forking a child and exiting from the partent

nghiant2710:
@petrosagg you're right. I've checked and updated the PR (please check my first comment)

imrehg:
@petrosagg @nghiant2710 ping on this issue ^^

alpine/aarch64/3.5/entry.sh
```diff
@@ -44,6 +44,12 @@ function init_openrc()
 	printf '%q ' "$@" >> /etc/resinApp.sh
 	chmod +x /etc/resinApp.sh
 
+	# Make sure there is no pidfile or process running before starting.
+	if [ -f "/var/run/resinapp.pid" ]; then
+		kill $(cat /var/run/resinapp.pid) &> /dev/null
```
petrosagg: You don't have to try and kill this. The problem was most likely caused by a container restart where the pid file persisted in the writable layer. So you can just leave the rm here

nghiant2710: Updated!

pimterry:
@petrosagg @nghiant2710 ping on this issue, we've got [a support thread](https://app.frontapp.com/open/cnv_6s7pxj) waiting on updates on it, looks like it's waiting for rereview.

PR #342, opened 2017-09-22
Fix #323, fix issue of broken systemd (/dev/console) in docker container

### LINKED ISSUE #323: Broken systemd in Debian Stretch and Fedora 25
Systemd v232 (Debian Stretch) and Systemd v231 (Fedora 25) are broken in docker container (ref https://github.com/moby/moby/issues/27202, https://github.com/systemd/systemd/pull/4262, https://bugzilla.redhat.com/show_bug.cgi?id=1373780#c4) Systemd can start but no logs from journald or any services. 
Since we don't have a patch to fix this issue yet, we should downgrade systemd to a working version.

nghiant2710:
connects to #323. We will create a process that always points to /dev/console to keep it opening.


petrosagg:
The commit seems to have unrelated changes in it. Please split into atomic changes

nghiant2710:
@petrosagg updated!

petrosagg:
@nghiant2710 I don't understand why systemd 230 needs all those changes. You need to be more descriptive in your commits

fedora/Dockerfile.armv7hf.tpl
```diff
@@ -33,26 +33,24 @@ RUN curl -SLO "http://resin-packages.s3.amazonaws.com/tini/v$TINI_VERSION/#{TINI
...
-        systemd-logind.service \
-        systemd-remount-fs.service \
-        getty.target \
-        graphical.target \
-        console-getty.service \
-        systemd-vconsole-setup.service
-
-COPY entry.sh /usr/bin/
+# We only want few core services run in the container.
+RUN find /etc/systemd/system \
```
petrosagg: We should use `systemctl enable/disable/mask` instead of deleting the files

fedora/Dockerfile.armv7hf.tpl
```diff
@@ -33,26 +33,24 @@ RUN curl -SLO "http://resin-packages.s3.amazonaws.com/tini/v$TINI_VERSION/#{TINI
...
+        -not -name '*udevd*' \
+        -not -name '*systemd-tmpfiles*' \
+        -not -name '*systemd-user-sessions*' \
+        -exec rm \{} \;
+
+COPY entry.sh /usr/bin/entry.sh
 COPY launch.service /etc/systemd/system/launch.service
 
-RUN systemctl enable launch.service systemd-udevd
+RUN systemctl enable /etc/systemd/system/launch.service
```
petrosagg: You don't need the full path of the service here

fedora/generate-dockerfile.sh
```diff
@@ -55,17 +55,29 @@ for repo in $repos; do
 	tiniChecksum="TINI_$repo" && tiniChecksum=$(eval echo \$$tiniChecksum)
 
 	for suite in $suites; do
+
+		if [ $suite == '24' ]; then
+			cgroup='VOLUME ["/sys/fs/cgroup"]'
+			cgroupEntry='mount -t tmpfs -o mode=0755 cgroup /sys/fs/cgroup'
```
petrosagg: Why is the cgroup not needed in the other case?

fedora/Dockerfile.armv7hf.tpl
nghiant2710: @petrosagg, if we don't delete those files, the logs will be very ugly and can make user confused. For example:
```
18.10.17 02:45:35 (+0700) Failed to install release agent, ignoring: No such file or directory
18.10.17 02:45:35 (+0700) systemd-remount-fs.service: Cannot add dependency job, ignoring: Unit systemd-remount-fs.service is masked.
18.10.17 02:45:35 (+0700) sys-kernel-config.mount: Cannot add dependency job, ignoring: Unit sys-kernel-config.mount is masked.
18.10.17 02:45:35 (+0700) dev-hugepages.mount: Cannot add dependency job, ignoring: Unit dev-hugepages.mount ismasked.
18.10.17 02:45:35 (+0700) systemd-modules-load.service: Cannot add dependency job, ignoring: Unit systemd-modules-load.service is masked.
18.10.17 02:45:35 (+0700) sys-fs-fuse-connections.mount: Cannot add dependency job, ignoring: Unit sys-fs-fuse-connections.mount is masked.
18.10.17 02:45:35 (+0700) kmod-static-nodes.service: Cannot add dependency job, ignoring: Unit kmod-static-nodes.service is masked.
18.10.17 02:45:35 (+0700) getty.target: Cannot add dependency job, ignoring: Unit getty.target is masked.
18.10.17 02:45:35 (+0700) systemd-logind.service: Cannot add dependency job, ignoring: Unit systemd-logind.service is masked.
18.10.17 02:45:35 (+0700) display-manager.service: Cannot add dependency job, ignoring: Unit display-manager.service is masked.
```

fedora/generate-dockerfile.sh
nghiant2710: We don't need to mount cgroup in some systemd latest versions, it relates to this issue https://github.com/systemd/systemd/issues/6477.

fedora/Dockerfile.armv7hf.tpl
petrosagg: @nghiant2710 why don't you disable instead of masking then?

nghiant2710: @petrosagg there're some static services which can't be disabled and they will output errors when systemd starts. 
I've updated this, we should keep masking services as we do now since the above errors seems to be systemd issue (I see the same issue here https://github.com/systemd/systemd/issues/5358). We can explain to users if they ask about these messages.

fedora/Dockerfile.armv7hf.tpl
nghiant2710: updated

PR #346, opened 2017-10-13
Update launch.Service to restart the service.

c00w:
Without this - if the binary exits and you're running a systemd init system, it will just hang - not getting restarted by docker or systemd.


nghiant2710:
LGTM, @petrosagg you good?

c00w:
I don't have write access, so someone else is going to need to merge this PR.

nghiant2710:
merged @c00w 👍

PR #350, opened 2017-10-19
 Add support for new device: Siemens IOT2000

nghiant2710:


device-base/generate-dockerfile.sh
```diff
@@ -102,6 +102,14 @@ for device in $devices; do
 		alpine_template='Dockerfile.alpine.tpl'
 		alpine_baseImage='i386-alpine'
 	;;
+	'IOT2000')
+		template='Dockerfile.tpl'
+		baseImage='i386-debian'
+		# TODO: Can't compile mraa on alpine linux atm, lack of necessary libraries.
+		#alpine_template='Dockerfile.alpine.i386.intel-edison.tpl'
+		alpine_template='Dockerfile.alpine.tpl'
```
floion: Does the above comment still stand?

nghiant2710: it's not related, just copy-paste issue. I removed the comments

PR #352, opened 2017-10-24
 Integrate the LOCK prefix fix for Iot2000 base images

nghiant2710:


device-base/Dockerfile.i386.iot2000.partial
```diff
@@ -0,0 +1,15 @@
+RUN set -x \
+	&& buildDeps='git-core autoconf libtool automake build-essential debhelper fakeroot cmake dpkg-dev devscripts' \
+	&& apt-get update && apt-get install -y $buildDeps --no-install-recommends && rm -rf /var/lib/apt/lists/* \
+	&& git clone https://github.com/mdr78/libx1000.git \
```
petrosagg: let's fix this to a specific commit

nghiant2710: @petrosagg, updated. It's the latest commit.

Issue #345, opened 2017-10-10
Duplicate tag in generated library file


nghiant2710:
There are cases when distro suite is the same with software base version (for example: Alpine Linux 3.5 and Python 3.5.x) which cause duplicate tag since the tag is from all possible combinations of distro suite and software version.

PR #357, opened 2017-11-01
Fix library generating script to make sure there will be no duplicate tag.

nghiant2710:
Connects to https://github.com/resin-io-library/base-images/issues/345.

An example of tag duplication is Python version 3.5.x and Alpine Linux 3.5, so tag `3.5` can point to an image with Python version 3.5.x on Alpine 3.6 or Python version 2.7.x on Alpine 3.5 since `2.7.x` is the latest version and `3.5` is a variant of all Alpine Linux Python images.

This PR will make sure that there is no duplication and language version will be prioritized, so tag `3.5` will point to Python version 3.5.x on Alpine 3.6.

### DIFF
--- a/golang/generate-stackbrew-library.sh
+++ b/golang/generate-stackbrew-library.sh
@@ -87,15 +87,15 @@ for device in "${devices[@]}"; do
 	for distro in "${distros[@]}"; do
 		# Debian
 		if [ $distro == 'debian' ]; then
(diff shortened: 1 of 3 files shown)


petrosagg:
I think we shouldn't support bare distro version tags at all. i.e a tag should either be `<language_version>` or `<language_version>-<distro_version>`. Never just `<distro_version>`. Otherwise we're just polluting the namespace

PR #371, opened 2017-12-19
 Install git to the base images for devices.

nghiant2710:
There are cases when users use `-slim` variant images have their builds broken because of git is missing. Also some tools like (pip) use git as well so essential package like git should be installed in the base images for devices.


petrosagg:
The goal of slim images is to be as lightweight as possible. Is this connected to some issue? Where did this come up?

nghiant2710:
@petrosagg Package manager tools like (npm, pip) require git to pull the packages. I've seen many users failed to install the resin-cli or other packages on slim base images because of missing git. So I think we should have it as a default package in all base images.

PR #372, opened 2018-01-11
Add gawk dependency

bbinet:
awk is required by `/usr/bin/entry.sh` entrypoint:
https://github.com/resin-io-library/base-images/blob/master/debian/entry.sh#L108

This should fix my "awk: command not found" issue when using `resin/armv7hf-debian:jessie` image:
```
root@hlnew:~# balena logs resin_supervisor
/usr/bin/entry.sh: line 108: awk: command not found
```

### DIFF
--- a/debian/Dockerfile.tpl
+++ b/debian/Dockerfile.tpl
@@ -24,6 +24,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		netbase \
 		curl \
 		udev \
+		gawk \
 	&& rm -rf /var/lib/apt/lists/*
 
 # Tini


nghiant2710:
@bbinet how did you build the base image for `resin_supervisor` container? 
Since i tested with `resin/armv7hf-debian:jessie` and `awk` is installed.
```
docker run -ti --rm resin/armv7hf-debian:jessie bash

root@83d19b8679a6:/# which awk
/usr/bin/awk
```

bbinet:
@nghiant2710 I was building my own resin_supervisor image, and it seems this was located in my build system (maybe a cache issue).
But everything is fine now.
Sorry for the noise.

PR #373, opened 2018-01-12
Yarn integration

### LINKED ISSUE #206: Consider providing yarn as part of the base image
https://yarnpkg.com/ is a modern npm client replacement from Facebook, Google, et. al.
Providing it pre-packaged is a matter of `npm install -g`, but should make our images look way more modern.
Would probably only make sense for `node >= 4`

nghiant2710:
Fixes #206, yarn has been integrated into the upstream base images for a while so we should have it in our base images too.

```
root@a42bb16e1fce:/opt# du -sh yarn
3.9M    yarn
```
This is the package size on amd64 base images so I think it will no longer be an issue to us anymore.


node/Dockerfile.alpine.slim.tpl
```diff
@@ -2,17 +2,33 @@
...
 
 # Install dependencies
 RUN apk add --no-cache libgcc libstdc++ libuv libcrypto1.0 libssl1.0
 
 RUN buildDeps='curl' \
 	&& set -x \
+	&& for key in \
+	6A010C5166006599AA17F08146C2130DFD2497F5 \
+	; do \
+		gpg --keyserver pgp.mit.edu --recv-keys "$key" || \
```
petrosagg: Why do you fetch the key from multiple key servers?

nghiant2710: @petrosagg, I get this from upstream dockerfile, I think they do that since a key server might not be available all the time. It happens a lot in our builds when a build fails due to a key server was unreachable

PR #376, opened 2018-01-18
Add VersionBot changes to repository (Autogenerate changelog only)

nghiant2710:
Change-Type: patch

### DIFF
--- a/versionist.conf.js
+++ b/versionist.conf.js
@@ -1,0 +1,67 @@
+'use strict';
+
+const execSync = require('child_process').execSync;
+
+const getAuthor = (commitHash) => {
+  return execSync(`git show --quiet --format="%an" ${commitHash}`, {
+    encoding: 'utf8'
+  }).replace('\n', '');
+};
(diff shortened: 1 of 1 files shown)


nghiant2710:
@hedss, can you review this PR and enable version bot for this repo please?

hedss:
@nghiant2710 Sure. One question, where did the `versionist` config for this come from? It looks like it's missing some options.

nghiant2710:
@hedss, I made this config file myself, basically I only want it to automatically update the changelog since we don't use version for this repo so please fix me if i'm wrong.

hedss:
@nghiant2710 It's missing some of the code used for checking tags. Several repos use a config that only updates Changelogs. You should be able to use https://github.com/resin-io/resin-router-service/blob/master/versionist.conf.js and just drop it in instead! :)

nghiant2710:
@hedss thanks. I made a small change in the template (version removed) since we dont need it here.

Issue #385, opened 2018-02-09
Add env var to disable/enable udev


shaunmulligan:
@nghiant2710 can we get this urgently

PR #386, opened 2018-02-19
Add UDEV env var to enable/disable udev for base images

nghiant2710:
Fixes https://github.com/resin-io-library/base-images/issues/385.
This PR will introduce a new base images env var `UDEV` (default to `on`) which will be used to enable/disable udev.

Notice: in case systemd is enabled and udev is disabled, the systemd udev service `systemd-udevd` will be masked (since it's a dependency for other services so disabling it won't work) so we need to mention in our docs if users want to enable it, they will need to unmask the service.

### DIFF
--- a/debian/aarch64/buster/entry.sh
+++ b/debian/aarch64/buster/entry.sh
@@ -4,13 +4,21 @@ set -m
 
 function start_udev()
 {
-	which udevd
-	if [ $? == '0' ]; then
-		udevd --daemon &> /dev/null
+	if [ "$UDEV" == "on" ]; then
+		if [ "$INITSYSTEM" != "on" ]; then
+			which udevd
+			if [ $? == '0' ]; then
+				udevd --daemon &> /dev/null
+			else
+				/lib/systemd/systemd-udevd --daemon &> /dev/null
+			fi
+			udevadm trigger &> /dev/null
+		fi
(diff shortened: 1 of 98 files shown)


debian/aarch64/buster/entry.sh
```diff
@@ -4,13 +4,21 @@ set -m
...
+			if [ $? == '0' ]; then
+				udevd --daemon &> /dev/null
+			else
+				/lib/systemd/systemd-udevd --daemon &> /dev/null
+			fi
+			udevadm trigger &> /dev/null
+		fi
 	else
-		/lib/systemd/systemd-udevd --daemon &> /dev/null
+		if [ "$INITSYSTEM" == "on" ]; then
```
shaunmulligan: @nghiant2710 can we not have the `else` branch of this run `systemctl unmask systemd-udevd` ? because then people won't need to put it in their dockerfiles if they want to re-enable it?

nghiant2710: @shaunmulligan, udev service can cause the race condition between the host and the container. If users disable udev, I don't think we need to keep this running,

shaunmulligan: @nghiant2710 okay, I think that makes sense :+1:

PR #409, opened 2018-04-03
Fix obsolete debian packages

nghiant2710:
Some packages are obsolete in new Debian releases so dockerfile template needs to be updated a bit.

### DIFF
--- a/buildpack-deps/Dockerfile.tpl
+++ b/buildpack-deps/Dockerfile.tpl
@@ -12,7 +12,6 @@ RUN apt-get update && apt-get install -y \
 		libjpeg-dev \
 		libmagickcore-dev \
 		libmagickwand-dev \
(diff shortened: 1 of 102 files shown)


nghiant2710:
@petrosagg I agree we should avoid having logic in the dockerfile as much as we can but with current generation script and since we support many Debian releases, I will go for one template file with simple logic that works for all variants rather than have a template for each variant and a complicated generation script.

This problem will be solved when we migrate to contracts since we can easily specify template for each dockerfile without changing the generation script logic.

petrosagg:
@nghiant2710 ok, in this case can you change the logic to be "if package X exists in the package list,  install X, otherwise install Y" instead of "try to install package X, if it fails install package Y"?

nghiant2710:
@petrosagg updated

PR #410, opened 2018-04-03
Ubuntu Base Images

joshughes:
We needed ubuntu xenial base images. So I tried to adapt the Debain scripts for Ubuntu. Happy to correct anything that may be wrong here. 

Front logo [Front conversations](https://app.frontapp.com/open/top_akuj)


nghiant2710:
@joshughes thanks a lot for the request, It will take a while until we can add it to our base images since we need to create ubuntu image for each device we support and the language stacks (node, python, go...) So in the meantime, can you just build the image from these Dockerfiles and let us know if you have any issues. Of course, we will notify you when the Ubuntu base images added to our base images.

jemoster:
Thanks for posting this pull-request!  I'm able to use this to run Ubuntu on a Jetson TX2 (aarch64).

I did need to run `chmod +x ubuntu/aarch64/xenial/entry.sh` for the entrypoint to run correctly when the container ran.  Other `entry.sh` files in the repository (such as `debian/aarch64/jessie/entry.sh`) have executable permissions on the committed files.  The other `entry.sh` files in this pull-request likely have the same issue as I don't see the extra file permission being set.

Hope this helps!

nghiant2710:
Hi, @jemoster you should use this PR instead https://github.com/resin-io-library/base-images/pull/431.

nghiant2710:
@joshughes, I will close this PR since I've submitted https://github.com/resin-io-library/base-images/pull/431, we will merge it soon and let you know when the ubuntu base images are released. Thanks

PR #454, opened 2018-08-06
Add DotNet Core 2.1 base images

richbayliss:
Work in progress for now


dotnet/Dockerfile.tpl
```diff
@@ -0,0 +1,27 @@
+# AUTOGENERATED FILE
+FROM #{FROM}
+
+RUN [ "cross-build-start" ]
+
+RUN apt-get update && \
+    apt-get install -y curl libunwind8 gettext icu-devtools
```
nghiant2710: can you pass extra option `--no-install-recommends` to the `apt-get install` command?

nghiant2710: and `rm -rf /var/lib/apt/lists/*` as well

dotnet/Dockerfile.tpl
```diff
@@ -0,0 +1,27 @@
...
+
+RUN rm -rf /tmp/* /var/tmp/* && \
+    rm -rf /var/lib/apt/lists/* && \
+    rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin
+
+WORKDIR /tmp
+
+RUN curl -sSL -o dotnet.tar.gz #{DOTNET_BUNDLE}
```
nghiant2710: Can you add the shasum check to verify the file integrity here, like this https://github.com/resin-io-library/base-images/blob/master/node/raspberrypi3/debian/9.11/Dockerfile#L14-L17?

dotnet/Dockerfile.tpl
```diff
@@ -0,0 +1,27 @@
...
+
+RUN mkdir -p /opt/dotnet && \
+    tar zxf dotnet.tar.gz -C /opt/dotnet && \
+    ln -s /opt/dotnet/dotnet /usr/local/bin
+
+RUN rm -rf /tmp/*
```
nghiant2710: Can we merge this RUN instruction with 2 previous ones so they just one layer is created instead of three?

dotnet/generate-all-dockerfile.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+set -e
+
+out_dir=.
+
+devices='raspberry-pi raspberry-pi2 raspberrypi3 beaglebone edison nuc vab820-quad zc702-zynq7 odroid-c1 odroid-ux3 parallella-hdmi-resin nitrogen6x cubox-i ts4900 colibri-imx6 apalis-imx6'
```
nghiant2710: There are some missing devices here I think, this is the list of all supported devices https://github.com/resin-io-library/base-images/blob/master/node/generate-latest-dockerfile.sh#L47

dotnet/generate-all-dockerfile.sh
```diff
@@ -0,0 +1,110 @@
...
+          bundle="https://download.microsoft.com/download/1/f/7/1f7755c5-934d-4638-b89f-1f4ffa5afe89/dotnet-runtime-2.1.2-linux-$dotnet_arch.tar.gz"
+          version='2.1.2'
+        ;;
+        'aspnetcore')
+          bundle="https://download.microsoft.com/download/1/f/7/1f7755c5-934d-4638-b89f-1f4ffa5afe89/aspnetcore-runtime-2.1.2-linux-$dotnet_arch.tar.gz"
+          version='2.1.2'
+        ;;
+      esac
+
+      dockerfilePath=$out_dir/$device/$base/$variant-$version
```
nghiant2710: can we update the `dockerfilePath` to be consistent with other language stack?
So it should be `$out_dir/$device/$base/$version/$variant`

dotnet/generate-all-dockerfile.sh
richbayliss: I will use that. It would be good to have a central `targets.sh` to include so that it can be centrally managed 👍

dotnet/generate-all-dockerfile.sh
richbayliss: Sure

dotnet/Dockerfile.tpl
richbayliss: Sure

dotnet/Dockerfile.tpl
richbayliss: the `rm -rf` was in the next run command, so I merged them up

galvesribeiro:
Folks any progress on this? Its being almost an year... We're starting on Balena and one of the services we're going to run is built on .Net Core.

richbayliss:
@galvesribeiro we are planning some new images using our `balenalib` base images, and that PR should go through soon :+1: It is planned to target `AMD64` and `armv7hf` (with the 3.0 previews also having `AARCH64` support)

PR #463, opened 2018-09-07
 new-device: Add new devices stem-x86-32 and revpi-core-3 

nghiant2710:


nghiant2710:
@floion  can I have a review please?

PR #474, opened 2018-09-26
new-device: Add build script info for asus-tinker-board and asus-tinker-board-s

nghiant2710:


nghiant2710:
@floion  can I have a review please?

PR #476, opened 2018-10-07
stack: Add blueprint for the language stacks

nghiant2710:
Change-type: minor


scripts/blueprints/stack-device.yaml
```diff
@@ -0,0 +1,26 @@
...
+  requires:
+    - type: hw.device-type
+      arch: "{{children.arch.sw.slug}}"
+  template:
+    - name: stack-device
+      data: >
+        # AUTOGENERATED FILE
+
+        FROM resin/{{hw.device-type.slug}}-{{sw.os.slug}}:{{sw.os.version}}
+
```
jviotti: Why do we have two blank lines here and below? Can we keep just one?

PR #480, opened 2018-10-18
Balena base images release

nghiant2710:


nghiant2710:
Connects to https://github.com/resin-io/contracts/pull/49

scripts/blueprints/os-arch.yaml
```diff
@@ -1,21 +1,22 @@
...
   sw.os: 1
   arch.sw: 1
+  sw.stack-variant: 1
   sw.blob: '1+'
 output:
   filename: Dockerfile
   type: sw.docker-image.os-arch
-  slug: "{{children.arch.sw.slug}}-{{children.sw.os.slug}}:{{children.sw.os.version}}"
-  path: "{{children.sw.os.slug}}/{{children.arch.sw.slug}}/{{children.sw.os.version}}"
+  slug: "{{this.children.arch.sw.slug}}-{{this.children.sw.os.slug}}"
```
jviotti: So we're not storing the version in the slug anymore? (i.e. after the `:`)? I see that we still have it on the path. Will the deploy scripts handle it?

scripts/blueprints/stack-arch.yaml
```diff
@@ -0,0 +1,25 @@
+selector:
+  sw.os: 1
+  sw.stack: 1
+  sw.stack-variant: 1
+  arch.sw: 1
+output:
+  filename: Dockerfile
+  type: sw.docker-image.stack-arch.{{this.children.sw.stack.slug}}
+  slug: "{{this.children.arch.sw.slug}}-{{this.children.sw.os.slug}}-{{this.children.sw.stack.slug}}"
```
jviotti: The slug is meant to be a unique identifier for the contract. Right now two images with different variants would contain the same slug, which feels wrong. I understand we are using the `slug` for the `FROM` part of the image, but maybe we can include variants in the slug, and have another contract property (i.e. `imageName`) that doesn't include it? 

Not an urgent thing right now, but it will make it easier when we start pushing these contracts to Jellyfish.

scripts/blueprints/stack-device.yaml
```diff
@@ -0,0 +1,26 @@
...
+  sw.os: 1
+  hw.device-type: 1
+  sw.stack: 1
+  sw.stack-variant: 1
+  arch.sw: 1
+output:
+  filename: Dockerfile
+  type: sw.docker-image.stack-device.{{this.children.sw.stack.slug}}
+  slug: "{{this.children.hw.device-type.slug}}-{{this.children.sw.os.slug}}-{{this.children.sw.stack.slug}}"
+  path: "{{this.children.sw.stack.slug}}/{{this.children.hw.device-type.slug}}/{{this.children.sw.os.slug}}/{{this.children.sw.os.version}}/{{this.children.sw.stack.version}}/{{this.children.sw.stack-variant.slug}}"
```
jviotti: Should we include the stack version in the path? Otherwise if there are stacks A v1 and A v2, the second one will override the first, right?

scripts/blueprints/os-arch.yaml
nghiant2710: @jviotti, this script will handle that task https://github.com/resin-io-library/base-images/pull/480/files#diff-ec6a48beb8b00bd417d941a15b908c8e. It will generate all possible tags for a specific image in which will be used as input for the deploy scripts.

scripts/blueprints/stack-arch.yaml
nghiant2710: @jviotti, I added `imageName` property in the blueprints and updated the slug so it will have stack versions and variants included. Hence the slug will always be unique.

scripts/blueprints/stack-device.yaml
nghiant2710: It's already there `{{this.children.sw.stack.slug}}/{{this.children.hw.device-type.slug}}/{{this.children.sw.os.slug}}/{{this.children.sw.os.version}}/**{{this.children.sw.stack.version}}**/{{this.children.sw.stack-variant.slug}}`

PR #487, opened 2018-11-02
stackbrew: Output library file in new format

nghiant2710:
Output all tags for a Dockerfile in a JSON to fix long build time issue

### DIFF
--- a/scripts/generate-stackbrew-library.js
+++ b/scripts/generate-stackbrew-library.js
@@ -84,8 +84,6 @@ function generateCombinations (arr) {
 
 
 function generateOsArchLibrary (context) {
-  var content = ''
-
   const osVersions = [context.children.sw.os.version]
   if (context.children.sw.os.version === context.children.sw.os.data.latest) {
     osVersions.push('latest')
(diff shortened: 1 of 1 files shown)


scripts/generate-stackbrew-library.js
```diff
@@ -106,17 +104,17 @@ function generateOsArchLibrary (context) {
     context.imageName
   )
 
-  tags.forEach((tag) => {
-    content += `${tag}: ${URL}@${commit.slice(0, -1)} ${path.join(ROOT, context.path)}\n`
-  })
+  const content = {
+    tag: tags[0],
+    repoDir: `${URL}@${commit.slice(0, -1)} ${path.join(ROOT, context.path)}`,
+    alias: tags.join(' ')
```
petrosagg: You should leave the tags as an array here and use `jq` in bashbrew to iterate on them

scripts/generate-stackbrew-library.js
```diff
@@ -143,11 +141,13 @@ function generateStackLibrary (context) {
...
```
petrosagg: same here

scripts/generate-stackbrew-library.js
nghiant2710: @petrosagg, I don't want to add more unnecessary complex logic to the bashbrew script since image tag only contains any letter, dot or dash so put all the alias tags in a string with space delimiter then we can iterate on them simple as this https://github.com/balena-io-library/official-images/pull/235/files#diff-015d29779976b02f9d9008807a990f07R72

petrosagg: I don't think it's unecessary logic nor that it's complex.

Actually I just realised you're using string functions to "parse" the JSON like `firstTag="${repoTag#*:}"`. This is not the right way to do thing, even if in this particular occasion it works so we should fix those too.

It's very simple to use `jq` from a bash script to properly interact with structured data and it doesn't make sense to compromise the data representation in order to use string functions to parse it. For example to get the repoDir all you need to do is `echo $json_data | jq .repoDir`

nghiant2710: @petrosagg we do use `jq` to parse the json first here https://github.com/balena-io-library/official-images/blob/08cecfd46f5e0fec2620c9f0a0a19cca2cdd0efb/bashbrew/bashbrew.sh#L313-L315. `firstTag="${repoTag#*:}"` as you see is just string manipulation, not JSON parsing.

PR #499, opened 2018-11-19
udev: Change how UDEV env var works

nghiant2710:
If Udev is set, check if current container is in privileged mode, if yes then mount devtmpfs and run udev, if no then a warning is printed.

### DIFF
--- a/scripts/assets/entry-alpine.sh
+++ b/scripts/assets/entry-alpine.sh
@@ -1,18 +1,51 @@
 #!/bin/bash
 
+hostname "$HOSTNAME" &> /dev/null
+if [[ $? == 0 ]]; then
+	PRIVILEGED=true
+else
+	PRIVILEGED=false
+fi
+
+function mount_dev()
+{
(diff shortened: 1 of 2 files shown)


scripts/assets/entry-alpine.sh
```diff
@@ -1,18 +1,29 @@
 #!/bin/bash
 
+hostname "$HOSTNAME" &> /dev/null
```
shaunmulligan: is there a reason we are bringing back the hostname setting?

scripts/assets/entry-alpine.sh
```diff
@@ -1,18 +1,29 @@
...
+if [[ $? == 0 ]]; then
```
shaunmulligan: @nghiant2710 what is the mechanism we are using to determine if the container is privileged or not?

nghiant2710: Run hostname command and check its exit code is the mechanism we use here

PR #503, opened 2018-11-29
docs: Add balenalib base images reference

nghiant2710:
Change-type: patch


nghiant2710:
`{{ $names.company.short }}` is now replaced with `balenalib`, however `{{ $links.githubMain }}` is not replaced yet, I guess we have that property available, right @shaunmulligan?

I will update this reference whenever we add a new device or new images so you can just fetch this file and use in main docs repo.

PR #573, opened 2019-08-03
entryscript: Change how entry script detects privileged container

nghiant2710:
Fixes https://jel.ly.fish/f5b9dd78-235c-4c93-ba1a-6a6691f48650

Since user can set value for HOSTNAME env var, if the value is not a valid host name then the entry script will always return it as unprivileged container.

Change-type: patch

### DIFF
--- a/examples/INITSYSTEM/openrc/entry.sh
+++ b/examples/INITSYSTEM/openrc/entry.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
 
-if hostname "$HOSTNAME" &> /dev/null; then
+# This command only works in privileged container
+ip link add dummy0 type dummy &> /dev/null
+if [[ $? == 0 ]]; then
 	PRIVILEGED=true
+	# clean the dummy0 link
+    ip link delete dummy0 &> /dev/null
 else
 	PRIVILEGED=false
 fi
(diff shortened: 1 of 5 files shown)


examples/INITSYSTEM/openrc/entry.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 
-if hostname "$HOSTNAME" &> /dev/null; then
+# This command only works in privileged container
+ip link add dummy0 type dummy &> /dev/null
+if [[ $? == 0 ]]; then
 	PRIVILEGED=true
+	# clean the dummy0 link
+    ip link delete dummy0 &> /dev/null
```
imrehg: Indentation seems to be off, compared to the line above.

examples/INITSYSTEM/systemd/systemd.v230/entry.sh
```diff
@@ -2,8 +2,12 @@
 
 set -m
 
-if hostname "$HOSTNAME" &> /dev/null; then
+# This command only works in privileged container
+ip link add dummy0 type dummy &> /dev/null
+if [[ $? == 0 ]]; then
 	PRIVILEGED=true
+	# clean the dummy0 link
+    ip link delete dummy0 &> /dev/null
```
imrehg: Indentation seems to be off compared to the line above?

PR #578, opened 2019-08-21
entryscript: test command directly rather than checking exit code

xginn8:
Change-type: patch
Signed-off-by: Matthew McGinn <matthew@balena.io>


nghiant2710:
@xginn8 LGTM, can you please apply on these entry scripts here as well https://github.com/balena-io-library/base-images/tree/master/scripts/assets?

xginn8:
@nghiant2710 done, PTAL!

nghiant2710:
Thanks a lot, @xginn8. I will merge this now

PR #583, opened 2019-09-16
test: Add initial tests for balenalib base images

nghiant2710:
Change-type: minor

### DIFF
--- a/scripts/assets/tests/test-os.sh
+++ b/scripts/assets/tests/test-os.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -e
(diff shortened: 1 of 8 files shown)


scripts/assets/tests/test-os.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -ex
```
jviotti: I think you can remove `-x` here (maybe a debugging leftover?)

nghiant2710: Updated

PR #616, opened 2020-03-10
Update base images to add onbuild messages, balena-info, balena-idle commands

### LINKED ISSUE #612: Add custom idle command
A very common technique we use is to have a container idling. This could be either to troubleshoot an issue on an existing container or just as part of the development workflow when you don't want your container to exit but you are experimenting so don't have a proper `CMD` to keep it running.

The most common way of doing this is using this one-liner: 
- `CMD while : ; do echo "${MESSAGE=Idling...}"; sleep ${INTERVAL=600};`

See this project for more reference:
- https://github.com/balena-io-playground/balena-idling

It would be great to be able to achieve this with a custom command similar to [install_packages](https://github.com/balena-io-library/base-images/blob/master/balena-base-images/armv7hf/debian/stretch/run/Dockerfile#L26-L49). We could have for example `idle` or `balena_idle`, so that the one-liner turns into:
- `CMD balena_idle`

See this thread for a use case example: https://www.flowdock.com/app/rulemotion/balenalabs/threads/3YFfDbq9BJfcNlyzDpqR4mVJmo_

nghiant2710:
Connects to https://github.com/balena-io/contracts/pull/150
Fixes #612

### DIFF
--- a/scripts/assets/commands/balena-idle
+++ b/scripts/assets/commands/balena-idle
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -e
+
+usage() {
+    cat <<EOUSAGE
+A small script that runs an idle process to keep the container running, while periodically printing the log: "Idling..."
+
+Options:
+  --message    Message to print when ilding (Default to "Idling...").
+  --interval   Printing interval in second (Default to 600 seconds).
+  --silent     No idle message.
+
+EOUSAGE
+}
+
+message="Idling..."
+interval=600
+is_silent=
+
+# Args handling
+opts="$(getopt -o 'h?' --long 'message:,interval:,silent' -- "$@" || { usage >&2 && exit 1; })"
+eval set -- "$opts"
+
+while true; do
+    flag=$1
+    shift
+    case "$flag" in
+        --message) message="$1" && shift ;;
+        --interval) interval="$1" && shift ;;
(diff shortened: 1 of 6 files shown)


scripts/assets/commands/balena-idle
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+usage() {
+	cat <<EOUSAGE
+This script basically idles to keep the container running, while periodically printing to the logs that "Idling...".
```
shaunmulligan, line 7: Rephrase as " A small script that runs an idle process to keep the container running, while periodically printing the log: "Idling...""

scripts/assets/commands/balena-idle
```diff
@@ -0,0 +1,40 @@
...
+
+Options:
+  -m	Message to print when ilding (Default to "Idling...").
```
shaunmulligan, line 10: perhaps we should add an option like `--silent` which doesn't log.

scripts/assets/commands/balena-idle
nghiant2710: @shaunmulligan, updated.

scripts/assets/commands/balena-idle
nghiant2710: updated.

PR #655, opened 2020-09-03
balena-base-images: Add l4t-core workaround for all Jetson images

acostach:
This adds a workaround for installing a dependency
called nvidia-l4t-core from the public debs.

Change-type: patch
Signed-off-by: Alexandru Costache <alexandru@balena.io>


nghiant2710:
Thanks a lot @acostach but we will make the changes in contracts, no need to manually update the Dockerfile here so this can be closed.

I've seen the issue created by @shaunmulligan and will make the change accordingly.

acostach:
Thanks @nghiant2710 , closing

Issue #658, opened 2020-09-24
Feature: Allow running entrypoint without command


Timple:
Since we also have an entrypoint for ourselves, we run both the entrypoint provided by balena as our own.

However the entrypoint supplied gives an error if the command does not exist. Would it be possible to allow passing no command?
This would save us the effort of maintaining and updating a duplicate of your entry.sh file.

nghiant2710:
The PR was merge and deployed.

PR #659, opened 2020-09-24
Feature: Allow running entrypoint without command

Timple:
Solves #658


nghiant2710:
LGTM but we need approve from other colleague as well. Hey @shaunmulligan, what do you think if we add this to the entryscript?

I've not tested this in other OSes we support but I think the behaviour will be the same and it's safe to add it.

PR #665, opened 2020-10-30
Allow running entrypoint without command for Alpine Linux

nghiant2710:
Connects-to: #659 
Change-type: minor


nghiant2710:
@balena-ci I self-certify!

PR #704, opened 2021-06-25
Update base images tests

nghiant2710:
This PR contains the following changes:
- Skip os name test for debian sid.
- Update link for go get test.


nghiant2710:
@balena-ci I self-certify!

PR #707, opened 2021-06-30
Delete CODEOWNERS

nghiant2710:
Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #716, opened 2021-08-25
Temporarily remove go get test

nghiant2710:
As the Go example repo does not work properly on some recent Go versions so let drop it for now

Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #732, opened 2021-11-13
Update dotnet hello world test

nghiant2710:
The hello world message is different among versions (`Hello World!` or `Hello, World!`) so the test should be updated to match them all

Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #784, opened 2022-07-15
Update Jenkinfile

nghiant2710:
Run cleanup job on every three days instead of daily to make sure this won't kill long running jobs


nghiant2710:
@balena-ci I self-certify!

PR #789, opened 2022-09-06
entryscript: check if sys/kernel mount point already exists

nghiant2710:
Since the supervisor can bind mount the host OS /sys into the container if io.balena.features.sysfs is set
we need to check whether it exists or not before mounting in the base-images entryscript

Change-type: patch

### DIFF
--- a/scripts/assets/entry-alpine.sh
+++ b/scripts/assets/entry-alpine.sh
@@ -31,7 +31,15 @@ function mount_dev()
 	# /dev/ptmx point to its ptmx.
 	# ref: https://www.kernel.org/doc/Documentation/filesystems/devpts.txt
 	ln -sf /dev/pts/ptmx /dev/ptmx
-	mount -t debugfs nodev /sys/kernel/debug
+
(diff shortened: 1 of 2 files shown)


nghiant2710:
Many thanks to @maggie0002 for the fix and your suggestions, since both `udev` and `bash` are the default packages mentioned in our docs so we cannot just drop them as it can break customers images who expect these packages are installed by default.

I think it's better if we just check the mountpoint before mounting as you do in your script, that should be enough to get rid of warning logs about existed mountpoint.

maggie0002:
We could follow the format of the rest of the script with lowercase for local vars, and without braces:

```
 	# When using io.balena.features.sysfs the mount point will already exist.
 	sysfs_dir='/sys/kernel/debug'

 	if ! mountpoint -q "$sysfs_dir"; then
 		mount -t debugfs nodev "$sysfs_dir"
 	fi
```

I am reading and not testing, it would need to be tested out.

nghiant2710:
changed to lowercase.

nghiant2710:
@balena-ci I self-certify!

PR #792, opened 2022-09-23
library: Generate all alias versions for sw.stack except python

nghiant2710:
Since Python stack has some 3.x series supported so should not generate the only major alias version for all but only latest series

Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #809, opened 2023-02-16
Update contracts to v2.0.44

nghiant2710:
Change-type: patch


nghiant2710:
@balena-ci I self-certify!

PR #819, opened 2023-03-31
Bump contrato to v0.8.0

pipex:
This will allow the scripts to use less memory so they can run as GH
actions.

Change-type: patch


klutchell:
lgtm

PR #823, opened 2023-04-03
Remove balena-supervisor-base contract

pipex:
The balena-supervisor-base images have not been used for a while by the supervisor dockerfile. This removes the contract and related scripts to avoid generating images that won't be used.

This PR also remove the scripts/stackbrew-library folder which doesn't seem to serve any purpose.

Change-type: minor


klutchell:
lgtm

PR #824, opened 2023-04-03
Use nodejs 18 for generating dockerfiles

Page-:
Change-type: patch


klutchell:
We won't actually be able to test the benefits until https://github.com/balena-io-library/base-images/pull/818 is merged.

Page-:
LGTM

Page-:
The benefits are more about being able to rely on the features/etc of that version for other work @klutchell, although it's possible it'll also be faster by itself (but I doubt it'll be significant if at all)

PR #826, opened 2023-04-03
Generate dockerfiles/descriptions in parallel

Page-:
Depends on #824 and https://github.com/product-os/contrato/pull/61


Page-:
LGTM

PR #828, opened 2023-04-04
Iterate the contrato.query for a concurrency of 5

Page-:
This allows for more concurrent work whilst still limiting the number of contracts we process concurrently to avoid excessive memory usage

Change-type: patch


Page-:
LGTM

PR #836, opened 2023-04-05
Disable balena-os stackbrew library file generation

klutchell:
Change-type: patch
See: https://github.com/balena-io-library/base-images/pull/775


klutchell:
I self-certify!

PR #838, opened 2023-04-12
Generate a patch of modified image files rather than the entire dir

klutchell:
Also remove existing files before committing new ones. This ensures we only generate files that are still reflected in the contracts.

Change-type: minor


pipex:
LGTM

klutchell:
- count modified files with `git status --porcelain | wc -l`
- deleting entire balena-base-images directory for a clean slate removes `110835` files
- after rebuilding with `npm run all` we are net `-31379`
- after rebuilding with `npm run all-dockerhub` we are net `-29522`

I'm okay with this if you are @pipex. It seems like we should only need current files, and just committing on top of the existing directory is not the way to go.

pipex:
@klutchell my guess is that we might have removed some contracts at some point but the old dockerfiles were never cleaned. Can you see which are the dockerfiles that are getting removed? This might give us some peace of mind.

klutchell:
Yeah, I reproduced those numbers locally, but a cursory glance at 31k files doesn't tell me much.

pipex:
I'm ok with this then, this won't remove any docker images, and if it turns out we missed something, someone will complain and we can revert :)

PR #842, opened 2023-04-14
Revert allowing empty commits

klutchell:
Instead we can upload an empty artifact and ignore failures when applying patches.

Change-type: patch


klutchell:
I self-certify!

PR #845, opened 2023-05-02
Defer generating stackbrew library files

klutchell:
Instead the official-images repo will have it's own workflow and base-images submodule to keep things
in-sync.

Change-type: patch


klutchell:
I self-certify!

PR #868, opened 2023-07-12
Remove unnecessary dependencies

pipex:
Switches from `eslint` to `@balena/lint`, removes `js-combinatorics` as it is not used and updates a few extra dependencies.

Change-type: patch


pipex:
I self-certify!

PR #879, opened 2023-08-09
Generate new installation token for push

klutchell:
It's possible that the old one has expired by this point.

Change-type: patch


klutchell:
I self-certify!

PR #891, opened 2023-08-21
Remove duplicate raspberrypi aliases

klutchell:
Change-type: patch
Resolves: https://github.com/balena-io-library/base-images/issues/877


klutchell:
I self-certify!

PR #916, opened 2023-09-08
Generate docker bake files instead of bashbrew libraries

klutchell:
GitHub actions build matrices are limited to 256 jobs
so we can't have over 2k jobs in one workflow.

Instead, create a new github workflow for every base library
and create a matrix for each active tag variant.

This should allow infinite scaling of devices rather
than single workflows that are limited in length, and matrix size.

Change-type: minor


klutchell:
I self-certify!

PR #919, opened 2023-09-11
Delete docs/balenalib-base-images-reference.md

klutchell:
This file is 4 years out-of-date and the current docs are auto-generated via the SDK.

Change-type: patch


klutchell:
I self-certify!

PR #920, opened 2023-09-11
Generate all dockerfile and libraries on Renovate PRs

klutchell:
Renovate is one of the few apps with permissions to update workflow files, so it is best suited to generate all workflows as a postUpgradeCommand.

Change-type: patch


klutchell:
I self-certify!

PR #936, opened 2023-09-29
Split workflows by os-arch and os-device

klutchell:
This allows us to change the blueprints and targets of the workflows script.


klutchell:
I self-certify!

PR #944, opened 2023-10-02
Perform a new checkout of master on finalize

klutchell:
The cached credentials in the existing checkout from Flowzone do not have the permissions we need.

Change-type: patch


klutchell:
I self-certify!

PR #974, opened 2023-10-30
Automatically create DockerHub repositories if they are missing

klutchell:
This requires the full username/password credentials for a user with admin access to the balenalib namespace. API tokens do not have the required permissions.

Change-type: minor


klutchell:
Superseded by https://github.com/balena-io-library/base-images/pull/975

PR #975, opened 2023-11-02
Write missing DockerHub repositories to the GitHub step summary

klutchell:
Change-type: patch


klutchell:
Something weird is going on, the list of missing repos is way longer than I expected, and many of these device types are not new.
https://github.com/balena-io-library/base-images/actions/runs/6737643828?pr=975

klutchell:
I self-certify!

PR #1058, opened 2024-05-08
Generate one workflow per device or arch

klutchell:
This creates a much more managable list of
workflows, at the cost of reducing the granularity of dependency triggers.

In this configuration, all related os-arch and stack-arch combinations will need to complete before triggering any of the downstream os-device and stack-device jobs.

Change-type: minor


klutchell:
I self-certify!

PR #1059, opened 2024-05-08
Add a schedule to trigger os-arch and stack-arch jobs

klutchell:
Use concurrency group settings to ensure that only
one bake workflow, no matter the device type or arch,
can be running at a single time. The rest will be queued
by default, with a manual option to cancel in-progress.

Since each workflow can generate 100s of jobs once
the matrices are expanded, we need this concurrency
limit to remain within the 500 concurrent runner
maximum allowed by our enterprise plan.


klutchell:
I self-certify!

PR #1061, opened 2024-05-09
Allow the Bake all workflow to be triggered manually

klutchell:
Use the workflow run event to trigger listening
jobs rather than calling them directly.

Use the same concurrency group for all bake workflows, allowing for cancellation of all jobs in progress without starting new ones.

Change-type: patch


klutchell:
I self-certify!

PR #1062, opened 2024-05-09
Make cancel-in-progress an opt-in feature with clear explanation

klutchell:
Make downstream workflows conditional on the triggering workflow returning success or failure, not skipped or cancelled.

Change-type: patch


klutchell:
I self-certify!

PR #1101, opened 2024-08-19
Kyle/self hosted

klutchell:


scripts/blueprints/workflows/os-arch.yaml
```diff
@@ -71,19 +71,20 @@ output:
...
           # https://github.com/docker/setup-buildx-action
           - name: Setup buildx
-            uses: docker/setup-buildx-action@d70bba72b1f3fd22344832f00baa16ece964efeb # v3
+            id: setup-buildx
+            uses: docker/setup-buildx-action@988b5a0280414f521da01fcc63a27aeeb4b104db # v3.6.1
             with:
               driver-opts: network=host
               install: true
+          # https://github.com/docker/setup-qemu-action
+          - name: Setup QEMU
```
pipex, line 82: These changes just switch the ordering of the actions, right? Or am I missing something? What is the reasoning behind that??

klutchell: We do this in Flowzone as well, by running buildx setup first, we can check the outputs to see which platforms are natively supported. So the Setup QEMU step ends up looking like this with the new `if` condition added by the node script.
```yaml
          # https://github.com/docker/setup-qemu-action
          - name: Setup QEMU
            uses: docker/setup-qemu-action@49b3bc8e6bdd4a60e6116a5414239cba5943d3cf # v3.2.0
            with:
              platforms: all
              # https://hub.docker.com/r/tonistiigi/binfmt
              image: tonistiigi/binfmt:qemu-v8.0.4-33
            if: contains(steps.setup-buildx.outputs.platforms, 'linux/arm/v6') == false
```
The only reason we add the if condition in node and not the template is I don't know how to escape all those special characters in contract templates like this, while still substituting the arch.

pipex (approved):

PR #1114, opened 2024-09-27
CI: Update checkout settings

Page-:
Change-type: patch


klutchell:
I was going to get to this today, thanks for doing it @Page- !

klutchell (changes requested):
@Page- these changes need to be made to the [workflow templates](https://github.com/balena-io-library/base-images/tree/master/scripts/blueprints/workflows), or they will get overwritten next time the workflows are generated.

Page-:
@klutchell this PR already updates those blueprints doesn't it?

klutchell:
So it does, apologies, they were buried at the end of the 106 changed files.

For future reference the PR workflow will build the workflow files and you don't need to change them manually.

klutchell (approved):

PR #1183, opened 2025-05-07
Update the DockerHub description after publishing

klutchell:
Change-type: minor

This requires that the `BALENAIMAGES_USER` be a direct
`Owner` of the `balenalib` organization in DockerHub.

See: https://balena.fibery.io/Work/Improvement/update-balena-base-images-Docker-README's-to-mark-them-deprecated-2750

### DIFF
--- a/scripts/blueprints/workflows/os-arch.yaml
+++ b/scripts/blueprints/workflows/os-arch.yaml
@@ -108,3 +108,14 @@ output:
               targets: ${{ matrix.target }}
               push: ${{ inputs.no-push != true }}
               provenance: false
+          # https://github.com/peter-evans/dockerhub-description
+          - name: Update DockerHub Description
+            if: inputs.no-push != true
+            continue-on-error: true
+            uses: peter-evans/dockerhub-description@432a30c9e07499fd01da9f8a49f0faf9e0ca5b77 # v4.0.2
+            with:
+              username: ${{ secrets.BALENAIMAGES_USER }}
+              password: ${{ secrets.BALENAIMAGES_TOKEN }}
+              repository: "balenalib/{{this.children.arch.sw.slug}}-{{this.children.sw.os.slug}}"
+              readme-filepath: ./balena-base-images/docs/dockerhub/{{this.children.arch.sw.slug}}-{{this.children.sw.os.slug}}-full-description.md
+              short-description: "This image is part of the balena.io base image series for IoT devices."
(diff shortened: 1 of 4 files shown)


vipulgupta2048 (commented):
Screenshot 2025-05-08 at 12 47 30 AM

Suggesting a few things.

.github/workflows/bake-aarch64.yml
```diff
@@ -92,6 +92,15 @@ jobs:
...
           push: ${{ inputs.no-push != true }}
           provenance: false
+      - name: Update DockerHub Description
+        if: inputs.no-push != true
+        continue-on-error: true
+        uses: peter-evans/dockerhub-description@432a30c9e07499fd01da9f8a49f0faf9e0ca5b77
+        with:
+          username: ${{ secrets.BALENAIMAGES_USER }}
+          password: ${{ secrets.BALENAIMAGES_TOKEN }}
+          repository: aarch64-alpine
```
vipulgupta2048, line 102:
```suggestion
          repository: balenalib/aarch64-alpine
```

Full name would be needed

klutchell:
> Suggesting a few things.

Yeah, the action claims to work with a token with those permissions. I'll work with @LPapapo tomorrow to figure out if the new token was used or not.

.github/workflows/bake-aarch64.yml
```diff
@@ -92,6 +92,15 @@ jobs:
...
+          readme-filepath: ./balena-base-images/docs/dockerhub/aarch64-alpine-full-description.md
```
vipulgupta2048, line 103: Feature creep maybe, but could add `short-description` field to update repository descriptions easily too since this is only going to run once and update the description with a deprecation notice. 

Optional.

vipulgupta2048 (commented):
Screenshot 2025-05-08 at 12 47 30 AM

Failing atm

.github/workflows/bake-aarch64.yml
klutchell: Good catch, that actually explains the error!

.github/workflows/bake-aarch64.yml
klutchell: This will run more than once. I want to merge this change now so any images released in the next few weeks would get updated READMEs, with or without deprecation notices.

We should open a future PR to update the short description to a hardcoded string when we publish the blog post.

.github/workflows/bake-aarch64.yml
klutchell: Unfortunately we still have the same error
https://github.com/balena-io-library/base-images/actions/runs/14892091622

Will check if the token has been used or not

vipulgupta2048 (approved):

